PR #4, opened 2025-02-22
Update license check CI target

haydentherapper:
By default, the check can only run on pull requests. We can configure it to work by setting `base-ref`, but it's sufficient to have this as a required check on PR runs.

### DIFF
--- a/.github/workflows/license_check.yml
+++ b/.github/workflows/license_check.yml
@@ -14,7 +14,7 @@
 
 name: License Check
 
-on: [push, pull_request]
+on: [pull_request]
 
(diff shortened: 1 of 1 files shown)


cmurphy:
Why is this needed? This check runs on push on rekor v1 https://github.com/sigstore/rekor/blob/85c62b0c0fd23b5e2cf4559f53b552f26582555e/.github/workflows/verify.yml#L22

cmurphy:
I see the problem is the new dependency license check

cmurphy (approved):

Issue #9, opened 2025-02-24
Add checkpoint note signers


haydentherapper:
**Description**

Issue #8, opened 2025-02-24
Add HTTP handlers for upload API and tiles APIs, including Tessera library calls and configuration options


haydentherapper:
**Description**

cmurphy:
Closed via https://github.com/sigstore/rekor-tiles/pull/98

PR #62, opened 2025-02-25
Add package for Tessera integration

cmurphy:
#### Summary

This change adds the workflow to instantiate the tessera storage driver
to add an entry to the log.

The TransparencyLogEntry structure uses signed integers for the LogIndex
and TreeSize values, while tessera mostly returns unsigned integers for
these values. This change incorporates a conversion mechanism.

#### Documentation

Partial https://github.com/sigstore/rekor-tiles/issues/8
Fixes https://github.com/sigstore/rekor-tiles/issues/9


pkg/signer/signer.go
```diff
@@ -0,0 +1,45 @@
...
+
+func New(ctx context.Context, signer string, pass string) (signature.Signer, error) {
+	switch {
+	case slices.ContainsFunc(kms.SupportedProviders(),
+		func(s string) bool {
+			return strings.HasPrefix(signer, s)
+		}):
+		return kms.Get(ctx, signer, crypto.SHA256)
+	default:
+		return NewFile(signer, pass)
```
loosebazooka, line 43: could this result in some confusing error messages? Where we don't find an expected matching KMS provider but then the default is invoked and fails. Then our error message is something about the NewFile constructor failing?

pkg/tessera/tessera.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+type Storage struct {
+	origin     string
+	awaiter    *tessera.IntegrationAwaiter
+	addFn      tessera.AddFn
+	readTileFn client.TileFetcherFunc
+}
+
+func NewStorage(ctx context.Context, origin, bucket, spanner, keyPath, keyPass string) (*Storage, error) {
```
loosebazooka, line 40: Does it make sense for this be named according to the backend type? How would we add other backend types?

Could bucket and spanner be abstracted interfaces for leaf storage and metadata storage?

loosebazooka, line 40: Similarly, instead of accepting keyPath and keyPass, could it just accept the signer already instantiated? (then the code can be explicit about local key or kms)

pkg/signer/file.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright The Rekor Authors.
```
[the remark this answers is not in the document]
cmurphy: This was copied without modification from rekor v1 and I'm not sure what the rules are for changing the copyright. I think it wasn't caught by the linter because it's not by "The Sigstore Authors".

pkg/signer/file.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/sigstore/sigstore/pkg/signature"
+	"go.step.sm/crypto/pemutil"
+)
+
+// returns an file based signer and verify, used for spinning up local instances
+type File struct {
+	signature.SignerVerifier
+}
+
+func NewFile(keyPath, keyPass string) (*File, error) {
```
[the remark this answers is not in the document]
cmurphy: Yeah it's just copied from rekor https://github.com/sigstore/rekor/blob/main/pkg/signer/file.go

pkg/signer/signer.go
cmurphy: Yeah that is strange, this was also copypasted from rekor but probably worth fixing.

pkg/tessera/tessera.go
cmurphy: I think it still makes sense for this to be `Storage` and `NewStorage` because the object is going to work the same no matter what the backend is. I'll change this to accept a `tessera.Driver` instead of a bucket and spanner string so the caller just has to create and pass in the driver, and it will work the same.

cmurphy: Will do

cmurphy:
@jku 

> TransparencyLogEntry is used in places like Bundle as well so should remain compatible, right?

Good point, it would probably be a minefield to change it.

pkg/signer/signer.go
cmurphy: Actually no, I think this is by design. If it isn't a known KMS schema, there is no other way to interpret it except as a file path. It's also not really possible to definitively look at a string and determine if it's a file path, it could start with '/' or not, it could even just be '-' to indicate stdin (https://github.com/smallstep/crypto/blob/84f676bc58c42233dd7679c8530122f63ad8e740/internal/utils/io.go#L23). This is working in rekor v1, I think I'll leave it as is.

My rekor checkout was old and didn't include the new Tink provider, so I'll update that part.

jku (commented):
looks fine to me (although the tessera integration itself I can't competently evaluate), left a few questions.

WRT unit testing, can we leverage existing tests in sigstore/rekor or is the implementation so different that this makes no sense?

pkg/signer/signer.go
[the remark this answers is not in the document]
jku: I see you already discussed this in another thread.

pkg/signer/signer.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// these are imported to load the providers via init() calls
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/aws"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/azure"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/gcp"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/hashivault"
+)
+
+// New returns a Signer for the given KMS provider, Tink, or a private key file on disk.
+func New(ctx context.Context, signer, pass, tinkKEKURI, tinkKeysetPath string) (signature.Signer, error) {
```
loosebazooka, line 36: I'm still a little confused by this `New`.  There's too many optional parameters in this method, its a bit funky. Is this just so we pass CLI options directly into the new?

Anyway, since we're just copying things over, we should probably leave as is, I guess I can go back in later and mess with it.

loosebazooka (commented):
lgtm, will wait on jku's comments to be resolved.

pkg/signer/signer.go
cmurphy: Basically yeah:

https://github.com/sigstore/rekor/blob/c820fcaf3afdc91f0acf6824d55c1ac7df249df1/pkg/api/api.go#L173-L178
https://github.com/sigstore/rekor/blob/c820fcaf3afdc91f0acf6824d55c1ac7df249df1/pkg/sharding/ranges.go#L149-L150

It got more complicated when tink was added in https://github.com/sigstore/rekor/pull/2228

pkg/tessera/note.go
```diff
@@ -0,0 +1,91 @@
...
+// noteSigner uses an arbitrary signer to sign messages in the sumdb/notes format.
+type noteSigner struct {
+	name string
+	hash uint32
+	sign func(msg []byte) ([]byte, error)
+}
+
+// Name returns the server name associated with the key.
+func (n *noteSigner) Name() string {
+	return n.name
```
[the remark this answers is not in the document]
cmurphy: I'll add some comments explaining. These methods are required so that `noteSigner` can implement https://pkg.go.dev/golang.org/x/mod/sumdb/note#Signer which Tessera requires for signing checkpoints.

pkg/tessera/note.go
```diff
@@ -0,0 +1,91 @@
...
+	return n.hash
+}
+
+// Sign returns a signature for the given message.
+func (n *noteSigner) Sign(msg []byte) ([]byte, error) {
+	return n.sign(msg)
+}
+
+// isValidName reports whether name is valid.
+// It must be non-empty and not have any Unicode spaces or pluses.
```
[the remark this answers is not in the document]
cmurphy: Here's the context https://github.com/C2SP/C2SP/blob/main/tlog-checkpoint.md#note-text I'll add a comment

pkg/tessera/safeint.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"math"
+)
+
+// safeInt64 holds equivalent int64 and uint64 integers.
+type safeInt64 struct {
+	u uint64
+	i int64
+}
```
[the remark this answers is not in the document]
cmurphy: It's not a guarantee by Tessera. It's an assumption in Rekor v1 and only because Rekor v1 uses signed ints. I'm using it here for compatibility with the TransparencyLogEntry definition, which uses signed ints. The operator would need to be aware of this limitation and start a new tree before the size reaches MaxInt64.

cmurphy (commented):
> WRT unit testing, can we leverage existing tests in sigstore/rekor or is the implementation so different that this makes no sense?

For this part, there's no equivalent in sigstore/rekor because there's no Tessera, and Rekor doesn't even directly use the TLE in its API responses. Once I actually start writing the HTTP handlers I may be able to pull more in from rekor v1.

cmurphy:
Here's the script I've been using to test this, if it's of interest https://gist.github.com/cmurphy/0ae387aceba197ce0490fa75d4e2ffd6

It also shows why I wanted to put pkg/signer in this PR even though pkg/tessera isn't directly using it.

pkg/tessera/note.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
+	"golang.org/x/mod/sumdb/note"
+)
+
+// noteSigner uses an arbitrary sigstore signer to implement golang.org/x/mod/sumdb/note.Signer,
+// which is used in Tessera to sign checkpoints in the signed notes format
+// (https://github.com/C2SP/C2SP/blob/main/signed-note.md).
+type noteSigner struct {
```
haydentherapper, line 40: Should we call this an ECDSA note signer? From [C2SP](https://github.com/C2SP/C2SP/blob/main/signed-note.md), the key ID calculation is for the `0x02` signature type. 

This is https://github.com/sigstore/rekor/issues/2062. 

We can use Tessera's note signer if the key is ed25519, and this signer if the key is ecdsa.

If it's RSA, we need to define our own key ID since that is not defined by C2SP. I'd suggest `SHA256(key name || 0xA || 0xFF || "RSA-PKCS#1v1.5" || public key)[:4]`. 0xFF comes from the spec, and we include an additional identifier recommended by the spec.

jku:
> lgtm, will wait on jku's comments to be resolved.

for the record, I consider them resolved: There is just no button to mark them as resolved.

cmurphy:
I split the signing part into a separate PR https://github.com/sigstore/rekor-tiles/pull/87 will rebase this when that is merged

pkg/tessera/safeint.go
haydentherapper, line 1: can you add a test for this?

pkg/tessera/tessera.go
```diff
@@ -0,0 +1,139 @@
...
+	hashes := make([][]byte, len(proof))
+	for i, h := range proof {
+		dst := make([]byte, hex.EncodedLen(len(h)))
+		hex.Encode(dst, h)
+		hashes[i] = dst
+	}
+	safeCheckpointSize, err := newSafeInt64(checkpoint.Size)
+	if err != nil {
+		return nil, fmt.Errorf("invalid tree size: %d", checkpoint.Size)
+	}
```
haydentherapper, line 123: What do you think about verifying the proof before returning it? I think it's a good practice just to double check something wasn't corrupted. We can later set up a metric for it and alert if it is ever incremented.

```
if err := proof.VerifyInclusion(rfc6962.DefaultHasher, idx.i, safeCheckpointSize, entry.LeafHash(), proof, checkpoint.Hash); err != nil { ... }
```

pkg/tessera/tessera.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("new proof builder: %w", err)
+	}
+	proof, err := proofBuilder.InclusionProof(ctx, idx.u)
+	if err != nil {
+		return nil, fmt.Errorf("generating inclusion proof: %w", err)
+	}
+	hashes := make([][]byte, len(proof))
+	for i, h := range proof {
+		dst := make([]byte, hex.EncodedLen(len(h)))
```
haydentherapper, line 116: I don’t think the hashes should be hex-encoded. 

The Rekor verifier expects hex encoded hashes https://github.com/sigstore/rekor/blob/26bfbf550e1c3b086851d0ccbf2d8ebf024c7f46/pkg/verify/verify.go#L141, but Sigstore-go will handle that - https://github.com/sigstore/sigstore-go/blob/797ca59e9032c12c7e4a39092eb91c9d7c1b3b63/pkg/tlog/entry.go#L118

We can write a new verifier library later on that doesn’t rely on the generated OpenAPI models.

pkg/tessera/tessera.go
cmurphy: I'm a little worried about the extra overhead this adds but I can add it for now and we can see if it becomes a problem later.

pkg/tessera/tessera.go
haydentherapper: Encoding bytes should be handled by the JSON marshaling library, as [base64](https://pkg.go.dev/encoding/json#Marshal:~:text=except%20that%20%5B%5Dbyte%20encodes%20as%20a%20base64%2Dencoded%20string).

IIRC Rekor v1 [hex-encoded bytes](https://github.com/sigstore/rekor/blob/26bfbf550e1c3b086851d0ccbf2d8ebf024c7f46/openapi.yaml#L632) to turn it into a string because OpenAPI 2 didn't properly handle bytes as return types.

Issue #7, opened 2025-02-24
Add Protobuf definitions for the two entry types and service endpoints, plus generated code


haydentherapper:
Our impl should roughly match what fulcio is doing here: https://github.com/sigstore/fulcio/tree/main/cmd/app

But we don't need a few things:
1. no legacy api (so we don't need a legacy fulcio client parallel)
2. duplex (grpc/http on same port)
3. potentially injecting the config into the context (which doesn't seem to be used?)

PR #64, opened 2025-02-27
Add docker continer based protoc builder

loosebazooka:
- Add a protoc-generator with isolated Makefile and go.mod
- Add a placeholder service definition
- Run the generator and check-in generated code

#### Summary

This doesn't actually add any useful protos which will be added in a followup, just the framework for code generation.
- source protos are in `api/protos`
- openapi is for documentation only, generated into `docs/openapi`
- generated go code is in `pkg/generated/protobuf/`
- all generation is in `protoc-builder`

#### Release Note

NONE

#### Documentation
Partial https://github.com/sigstore/rekor-tiles/issues/7


protoc-builder/Makefile
```diff
@@ -0,0 +1,76 @@
+#
+# Copyright 2022 The Sigstore Authors.
```
cmurphy, line 2: Wrong copyright year? Not sure why the linter didn't see this

cmurphy (commented):
> openapi is for documentation only

There's no equivalent for protobuf I guess?

loosebazooka:
> There's no equivalent for protobuf I guess?

both openapi and proto defs are in `./api/`,

protoc-builder/Makefile
loosebazooka: yeah I was wondering the same thing about addlicense, it seems to not actually check that.

cmurphy:
I think docs are good, I'm just curious if protobuf has some native way of generating docs or if the only option is openapi/swagger. If that's the only option then that's fine.

loosebazooka:
kk moved openapi out to docs

cmurphy (approved):

jku (approved):
I think it's worth it to have something running ASAP, but let's make sure there's an issue open to do something about build system here: the amount checksums and version numbers in the Makefile and Dockerfile does not say good things about long term maintainability.

PR #83, opened 2025-03-05
add more auto formatting

loosebazooka:
#### Summary
Add licenses and goimports to `make lint` target

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/.github/workflows/protoc_diff_check.yml
+++ b/.github/workflows/protoc_diff_check.yml
@@ -20,17 +20,11 @@ permissions:
 
 on:
   push:
-    paths:
-      - 'api/proto/**'
-      - 'docs/openapi/**'
-      - 'pkg/generated/protobuf/**'
-      - 'protoc-builder/**'
+    branches:
+      - main
   pull_request:
-    paths:
-      - 'api/proto/**'
-      - 'docs/openapi/**'
-      - 'pkg/generated/protobuf/**'
-      - 'protoc-builder/**'
+    branches:
+      - main
 
 jobs:
   check_generated_protos:
@@ -42,7 +36,7 @@ jobs:
       with:
         persist-credentials: false
 
-    - name: Clear out protobuf directories
+    - name: Clean
       run: |
         make clean
 
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: all test clean lint gosec ko-local
+.PHONY: all test clean lint gosec ko-local tools
 
 all: rekor-server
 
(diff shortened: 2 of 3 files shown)


haydentherapper (dismissed):

cmurphy (dismissed):

ramonpetgrave64 (changes requested):
We're changing the make command for generating protos, but the workflow to check for proto changes didn't run in this PR.

- https://github.com/sigstore/rekor-tiles/actions/workflows/protoc_diff_check.yml

I think it's okay to remove the `paths` specifier in the workflow file.

 - https://github.com/sigstore/rekor-tiles/blob/7d96f461d582ab77fccdd6c16ecf7ec63bf70159/.github/workflows/protoc_diff_check.yml#L23-L33

loosebazooka:
> but the workflow to check for proto changes didn't run in this PR., yeah my local repo was in a weird state

loosebazooka:
hrmm... some dependencies of the tools have incompatible licenses?

.github/workflows/protoc_diff_check.yml
```diff
@@ -20,17 +20,11 @@ permissions:
 
 on:
   push:
-    paths:
-      - 'api/proto/**'
-      - 'docs/openapi/**'
-      - 'pkg/generated/protobuf/**'
-      - 'protoc-builder/**'
+    branches:
```
haydentherapper, line 23: Do we need on push? Making this a required workflow on pull_request could be sufficient

Makefile
```diff
@@ -32,24 +32,27 @@ ifeq ($(DIFF), 1)
...
 
-GOBIN ?= $(shell go env GOPATH)/bin
-
-lint: ## Run golangci-lint checks
-	$(GOBIN)/golangci-lint run -v ./...
+GOBIN = $(abspath ./tools/bin)
 
-addlicense: ## Add licenses to source files
+lint: tools
 	$(GOBIN)/addlicense -l apache -c "The Sigstore Authors" -ignore "third_party/**" -v *
```
haydentherapper, line 48: to avoid GOBIN, you can also use `go tool <name>`  from https://tip.golang.org/doc/modules/managing-dependencies#tools

.github/workflows/protoc_diff_check.yml
loosebazooka: I guess not, maybe just main

Makefile
```diff
@@ -40,16 +44,15 @@ SERVER_LDFLAGS=$(REKOR_LDFLAGS)
 
 GOBIN ?= $(shell go env GOPATH)/bin
 
-lint: ## Run golangci-lint checks
-	$(GOBIN)/golangci-lint run -v ./...
-
-addlicense: ## Add licenses to source files
+lint: ## add licenses, cleanup imports/gofmt, run golangci-lint checks
 	$(GOBIN)/addlicense -l apache -c "The Sigstore Authors" -ignore "third_party/**" -v *
```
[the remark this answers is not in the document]
loosebazooka: just did a go install for now.

.github/workflows/protoc_diff_check.yml
ramonpetgrave64: I think we might need it, just in case if we allow merging without the PR content already being in sync with main.

haydentherapper (approved):

PR #87, opened 2025-03-07
Add note signers

cmurphy:
This change copies most of the sigstore/rekor/pkg/signer package to enable signing checkpoints with an on-disk key file or a KMS provider, and adds a note helper package to convert the Rekor signers to the sumdb/note signing format.

Fixes https://github.com/sigstore/rekor-tiles/issues/9
Relates to https://github.com/sigstore/rekor/issues/2062


cmurphy:
I've not tested this with tink or other KMSs yet.

pkg/note/note.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
+	"golang.org/x/mod/sumdb/note"
+)
+
+const (
+	algEd25519 = 1
+	algUndef   = 255
+	rsaID      = "RSA-PKCS#1v1.5"
```
haydentherapper, line 43: Can follow up later, but we should document this custom ID in a readme on log entry verification.

pkg/signer/signer.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// these are imported to load the providers via init() calls
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/aws"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/azure"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/gcp"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/hashivault"
+)
+
+// New returns a Signer for the given KMS provider, Tink, or a private key file on disk.
+func New(ctx context.Context, signer, pass, tinkKEKURI, tinkKeysetPath string) (signature.Signer, error) {
```
haydentherapper, line 38: I know this was copied in, but I (who wrote this) never loved passing a bunch of strings, prone to mixing them up. Do you want to change this to a config struct (or file a follow up issue)?

pkg/signer/signer.go
```diff
@@ -0,0 +1,50 @@
...
+	switch {
+	case slices.ContainsFunc(kms.SupportedProviders(),
+		func(s string) bool {
+			return strings.HasPrefix(signer, s)
+		}):
+		return kms.Get(ctx, signer, crypto.SHA256)
```
haydentherapper, line 44: (Also can be a follow up) Related to the crypto agility work we're doing now, we found in Fulcio that hardcoding this hash caused ECDSA P384 and P521 keys to not work correctly since those are typically associated with SHA-384 and SHA-512, especially for KMS providers.

Can we add an option to pass in the hash algorithm?

pkg/signer/file.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// NewFileSigner returns an file-based signer and verify, used for spinning up local instances.
+func NewFileSigner(keyPath, keyPass string) (*File, error) {
+	opaqueKey, err := pemutil.Read(keyPath, pemutil.WithPassword([]byte(keyPass)))
+	if err != nil {
+		return nil, fmt.Errorf("file: provide a valid signer, %s is not valid: %w", keyPath, err)
+	}
+
+	signer, err := signature.LoadSignerVerifier(opaqueKey, crypto.SHA256)
```
haydentherapper, line 41: (Also can do in a follow up) Can we change from a hardcoded hash to the typical signature-hash pairing in https://github.com/sigstore/sigstore/pull/2014/files#diff-90f2b413b0fbeae56caac0481e57d7dff1bde02810bce967bd3f8c7d98670010R242?

Any RSA key size uses SHA-256, ECDSA P256/P384/P521 uses SHA-256/384/512 respectively, and Ed25519 uses SHA-512 but that's harcoded into the implementation.

haydentherapper (approved):

pkg/note/note.go
```diff
@@ -0,0 +1,152 @@
...
+		keyID, err = rsaKeyHash(origin, pk)
+		if err != nil {
+			return &noteSigner{}, fmt.Errorf("getting RSA key hash: %w", err)
+		}
+	default:
+		return &noteSigner{}, fmt.Errorf("unsupported key type: %T", pubKey)
+	}
+
+	sign := func(msg []byte) ([]byte, error) {
+		return signer.SignMessage(bytes.NewReader(msg), options.WithContext(ctx))
```
loosebazooka, line 144: Could context be an input into the SignMessage function rather than the NewNoteSigner func?

pkg/note/note.go
```diff
@@ -0,0 +1,152 @@
...
+	return n.name
+}
+
+// KeyHash returns the key hash.
+func (n *noteSigner) KeyHash() uint32 {
+	return n.hash
+}
+
+// Sign returns a signature for the given message.
+func (n *noteSigner) Sign(msg []byte) ([]byte, error) {
```
loosebazooka, line 66: Hrmm, I guess there's not really a way to inject context here without modifying the interface?

pkg/note/note.go
cmurphy: https://github.com/sigstore/rekor-tiles/issues/93

pkg/note/note.go
cmurphy: Correct, this method signature cannot be changed.

pkg/note/note.go
cmurphy: It's not possible without breaking the note signer interface. I wouldn't expect the program to want to create a new context every time it signs something, it would typically use a long running server context.

pkg/signer/file.go
cmurphy: I can change this to `signature.LoadDefaultSignerVerifier` which was added ... today.

haydentherapper: SG, I'll cut a sigstore/sigstore release now

haydentherapper: https://github.com/sigstore/sigstore/releases/tag/v1.9.1

pkg/note/note.go
cmurphy: @haydentherapper what's the reasoning behind your suggestion to use `PKCS#1` as the identifier, vs `PKCS#8` or `PKIX`? It has implications on whether I use https://pkg.go.dev/crypto/x509#MarshalPKCS1PublicKey or https://pkg.go.dev/crypto/x509#MarshalPKIXPublicKey to convert the key object to bytes, I just noticed I'm using the wrong one for this identifier.

haydentherapper: To differentiate between the two modern RSA schemes, PSS and PKCS#1v1.5. You're right that there are two ways to encode public keys, so can we include both the scheme and public key encoding in the ID name?

haydentherapper: Including encoding also matches up well with how we’ve named the supported keys - https://github.com/sigstore/protobuf-specs/blob/e3f5847bde2c42348d1b99add0737ab3193f45d0/protos/sigstore_common.proto#L69

pkg/note/note.go
loosebazooka: Oh I was thinking that each signing event would be cancellable/timeoutable and that's why it would have a separate context config

cmurphy: If this was tied to requests then that would be desirable, but checkpoint signing happens in batches and isn't 1:1 with incoming requests. If I'm wrong and we do want the signing event to be cancellable independent of the main server context then we could always have multiple instances of the signer.

pkg/signer/tink.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	tinkUtils "github.com/sigstore/rekor/pkg/signer/tink"
```
haydentherapper, line 27: Can we pull in the Tink utilities to remove the dependency on rekor?

jku: it's a bit strange since it already looks like we do -- isn't that pkg/signer/tink/  in this PR?

pkg/note/note.go
loosebazooka: Ah I see, okay cool thanks for the clarification.

pkg/signer/tink.go
cmurphy: Yes, just forgot to update the import.

pkg/signer/tink/tink.go now exists in fulcio, rekor and rekor-tiles, maybe we should just put it in sigstore/sigstore?

haydentherapper: Yep, that sounds good.

haydentherapper (approved):

PR #82, opened 2025-03-04
WIP: Get server up and running

loosebazooka:

PR #92, opened 2025-03-10
Add grpc and http servers.

loosebazooka:
Starts up servers and makes them available on localhost

todos:
- Add optional tls protection when communicating between http proxy and grpc service (see fulcio impl)
- Forward command line options to config objects (we use defaults)

#### Summary
Final part of #82

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/pkg/server/serve_test.go
+++ b/pkg/server/serve_test.go
@@ -40,7 +40,7 @@ func TestServe_smoke(t *testing.T) {
 	slog.SetDefault(slog.New(slog.NewTextHandler(os.Stdout, nil)))
 	gc := NewGRPCConfig()
 	hc := NewHTTPConfig()
-	s := &testServer{}
+	s := &mockServer{}
 
 	// Start the server
 	var wg sync.WaitGroup
--- a/pkg/server/serve_test.go
+++ b/pkg/server/serve_test.go
@@ -1,0 +1,149 @@
+// Copyright 2025 The Sigstore Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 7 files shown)


pkg/server/service_test.go
haydentherapper, line 1: nit: might name this service_mock instead

pkg/server/serve_test.go
```diff
@@ -0,0 +1,149 @@
...
+	checkHTTPGet(t, httpBaseURL+"/api/v2/tile/1/2.p/3", "test-tile:1,2.p,3")
+	checkHTTPGet(t, httpBaseURL+"/api/v2/tile/entries/1", "test-entries:1")
+	checkHTTPGet(t, httpBaseURL+"/api/v2/tile/entries/1.p/2", "test-entries:1.p,2")
+
+	// Simulate SIGTERM to trigger graceful shutdown
+	if err = syscall.Kill(syscall.Getpid(), syscall.SIGTERM); err != nil {
+		t.Fatalf("Could not kill server")
+	}
+
+	wg.Wait()
```
haydentherapper, line 88: should we check that the server is offline after sending SIGTERM?

haydentherapper (dismissed):

pkg/server/serve_test.go
loosebazooka: I think the test doesn't end if the sigterm doesn't work

pkg/server/service_test.go
jku: +1, or "rekor_mock" even. Took me a moment to understand what it was for

haydentherapper:
Can you rename the file too, to differentiate it from unit tests?

loosebazooka:
> Can you rename the file too, to differentiate it from unit tests?

Ah I see, yeah

haydentherapper (approved):

PR #98, opened 2025-03-13
Add server config options and request handlers

cmurphy:
Add options for configuring the server, the storage backend, and the signing provider. Implement the entry upload request handler.


pkg/server/service.go
```diff
@@ -27,10 +29,16 @@ import (
...
 }
 
-func (s *Server) CreateEntry(context.Context, *pb.CreateEntryRequest) (*pbs.TransparencyLogEntry, error) {
-	return nil, status.Errorf(codes.Unimplemented, "method CreateEntry not implemented")
+func NewServer(storage *tessera.Storage) *Server {
+	return &Server{storage: storage}
+}
+
+func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*pbs.TransparencyLogEntry, error) {
+	// FIXME: translate pb.CreateEntryRequest to tessera.Entry
```
cmurphy, line 40: This will take longer than expected, a significant part of https://github.com/sigstore/rekor/tree/main/pkg/types needs to be reimplemented for the new protobuf object which works differently from the old types, e.g. there's no `Kind()` or `APIVersion()` any more.

pkg/types/hashedrekord/hashedrekord.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func Validate(hr *pb.HashedRekordRequest) error {
+	if len(hr.Signature) == 0 {
+		return fmt.Errorf("missing signature")
+	}
+	if hr.Verifier == nil {
+		return fmt.Errorf("missing verifier")
+	}
+	if hr.Verifier.GetPublicKey() == nil && hr.Verifier.GetX509Certificate() == nil {
```
cmurphy, line 30: Old hashedrekord didn't have separate fields for public key and x.509 cert https://github.com/sigstore/rekor/blob/4b01f831f8b23f9070f59f4e476e3eabc5b68ab5/pkg/generated/models/hashedrekord_v001_schema.go#L394

pkg/server/service.go
```diff
@@ -27,11 +34,48 @@ import (
...
+			return nil, fmt.Errorf("validating hashedrekord request: %w", err)
+		}
+		serialized, err = json.Marshal(hr)
+		if err != nil {
+			return nil, fmt.Errorf("parsing hashedrekord request: %w", err)
+		}
+	case *pb.CreateEntryRequest_DsseRequest:
+		ds := req.GetDsseRequest()
+		if err := dsse.Validate(ds); err != nil {
+			// FIXME(cmurphy): this needs to return a 4XX status code to the server, right now it's returning a 500
```
cmurphy, line 61: ?

cmd/rekor-server/app/serve.go
```diff
@@ -16,38 +16,130 @@
...
+	serveCmd.Flags().String("gcp.bucket", "", "GCS bucket for tile and checkpoint storage")
+	serveCmd.Flags().String("gcp.spanner", "", "Spanner database URI")
+
+	// checkpoint signing configs
+	serveCmd.Flags().String("signer.type", "file", "type of signer to use, must be one of 'file', 'kms', or 'tink'")
+	serveCmd.Flags().String("signer.filepath", "", "path to the signing key")
+	serveCmd.Flags().String("signer.password", "", "password to decrypt the signing key")
+	serveCmd.Flags().String("signer.kmskey", "", "URI of the KMS key, in the form of awskms://keyname, azurekms://keyname, gcpkms://keyname, or hashivault://keyname")
+	serveCmd.Flags().String("signer.kmshash", "sha256", "hash algorithm used by the KMS")
+	serveCmd.Flags().String("signer.tink_kek_uri", "", "encryption key for decrypting Tink keyset. Valid options are [aws-kms://keyname, gcp-kms://keyname]")
```
haydentherapper, line 116: nit, preference for hyphenated CLI flags, following https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

cmd/rekor-server/app/serve.go
```diff
@@ -16,38 +16,130 @@
...
+	serveCmd.Flags().String("signer.tink_keyset_path", "", "path to encrypted Tink keyset")
+	if err := viper.BindPFlags(serveCmd.Flags()); err != nil {
+		slog.Error(err.Error())
+		os.Exit(1)
+	}
 	rootCmd.AddCommand(serveCmd)
 }
+
+var hashAlgMap = map[string]crypto.Hash{
```
haydentherapper, line 125: We can limit this to SHA256, SHA384, and SHA512.

From https://github.com/sigstore/protobuf-specs/blob/4752ba2fcab1501e4e901164f64ecad22b3586da/protos/sigstore_common.proto#L37, which includes SHA3 algs but I'm unaware of any KMS providers supporting SHA3 currently.

cmd/rekor-server/app/serve.go
```diff
@@ -16,38 +16,130 @@
...
+	// server configs
+	serveCmd.Flags().Int("http.port", 3000, "HTTP port to bind to")
+	serveCmd.Flags().String("http.address", "127.0.0.1", "HTTP address to bind to")
+	serveCmd.Flags().Int("grpc.port", 3001, "GRPC port to bind to")
+	serveCmd.Flags().String("grpc.address", "127.0.0.1", "GRPC address to bind to")
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "localhost"
+	}
+	serveCmd.Flags().String("hostname", hostname, "public hostname of instance")
```
haydentherapper, line 104: Can we mention this is used as part of the checkpoint?

pkg/types/dsse/dsse.go
```diff
@@ -0,0 +1,31 @@
...
+	"fmt"
+
+	pb "github.com/sigstore/rekor-tiles/pkg/generated/protobuf"
+)
+
+func Validate(ds *pb.DSSERequest) error {
+	if ds.Envelope == "" {
+		return fmt.Errorf("missing envelope")
+	}
+	if len(ds.Verifier) == 0 {
```
haydentherapper, line 27: Can we also add a function to validate verifiers contain non-empty `raw_bytes` and an identifier for a public key, or non-empty `raw_bytes` for a cert?

pkg/types/dsse/dsse.go
```diff
@@ -0,0 +1,31 @@
...
+		return fmt.Errorf("missing verifiers")
+	}
```
haydentherapper, line 29: We will also need to verify the signature over the payload for DSSE and the signature over the artifact hash for hashedrekord. Do we want to do that in Validate? Add a TODO to follow up later?

pkg/types/dsse/dsse.go
haydentherapper: Ah I see you started doing that in hashedrekord.go, we could extract that into a func to validate verifiers in both types.

pkg/types/hashedrekord/hashedrekord.go
```diff
@@ -0,0 +1,41 @@
...
+	// TODO(cmurphy): hash/public key compatibility check (https://github.com/sigstore/rekor-tiles/issues/100)
+	if hr.Verifier.GetPublicKey() == nil && hr.Verifier.GetX509Certificate() == nil {
+		return fmt.Errorf("missing signature public key or X.509 certificate")
+	}
+	if hr.Data == nil {
+		return fmt.Errorf("missing data")
+	}
+	if len(hr.Data.Digest) == 0 {
+		return fmt.Errorf("missing data digest")
+	}
```
haydentherapper, line 39: Check `hr.Data.HashAlgorithm` isn't `HASH_ALGORITHM_UNSPECIFIED`?

pkg/server/service.go
```diff
@@ -27,11 +34,56 @@ import (
...
+	var serialized []byte
+	var err error
+	switch req.GetSpec().(type) {
+	case *pb.CreateEntryRequest_HashedRekordRequest:
+		hr := req.GetHashedRekordRequest()
+		if err := hashedrekord.Validate(hr); err != nil {
+			slog.Warn("failed validating hashedrekord request", "error", err.Error())
+			return nil, status.Errorf(codes.InvalidArgument, "invalid hashedrekord request")
+		}
+		serialized, err = json.Marshal(hr)
```
haydentherapper, line 54: Do we need to use https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#Marshal?

cmurphy: I don't think so because `hr` is a regular struct not a `proto.Message`. @loosebazooka ?

pkg/types/dsse/dsse.go
cmurphy: Thanks for catching that, that validation was hidden in the unmarshal methods. Filed https://github.com/sigstore/rekor-tiles/issues/101

pkg/types/hashedrekord/hashedrekord.go
cmurphy: I think that will be covered in https://github.com/sigstore/rekor-tiles/issues/100

pkg/types/dsse/dsse.go
cmurphy: Moved to pkg/types/verifier

haydentherapper (dismissed):

pkg/server/service.go
haydentherapper: It looks like this does implement `proto.Message`, with the struct implementing [these methods](https://pkg.go.dev/google.golang.org/protobuf/runtime/protoiface#MessageV1). I think `json.Marshal` can work for some structs, but not if the struct contains any proto-specific messages like one-ofs.

cmurphy: Oops you're right, switching it

haydentherapper (dismissed):

pkg/types/verifier/verifier.go
```diff
@@ -0,0 +1,39 @@
...
+		return fmt.Errorf("missing signature public key or X.509 certificate")
+	}
+	if publicKey != nil {
+		if len(publicKey.GetRawBytes()) == 0 {
+			return fmt.Errorf("missing public key raw bytes")
+		}
+	}
+	if x509Cert != nil {
+		return fmt.Errorf("missing X.509 certificate raw bytes")
+	}
```
jku, line 37: This looks like it's missing some code

cmurphy: :woman_facepalming:

jku (approved):

cmurphy:
Added tests for hashedrekord, dsse, and verifier validation, and for the create entry handler. The handler test is not very in depth, as much of it is covered by the tessera Add test or relies on tessera internals, it will be covered more thoroughly in e2e tests.

haydentherapper (approved):

PR #107, opened 2025-03-17
Codecov tweaks

### LINKED ISSUE #104: coverage upload fails on main
`error - 2025-03-14 00:13:39,481 -- Upload failed: {"message":"Token required because branch is protected"}`

* `codecov/codecov-action` considers this a success because we don't explicitly set `fail_ci_if_error` -- I wonder if the action is unreliable if they set that to false by default?
* codecov runs on PRs seem to succeed but lack of baseline from main means the codecov comments in PRs are garbage

jku:
* Enable oidc auth for codecov -- use a workaround since the actual auth mechanism is broken
* This adds `id-token: write` permission in the unit-tests job. I think this is ok: test.yml impersonation is not a big issue
* Also add `fail_ci_if_error: true` so we actually notice if the action fails -- we can revert this if failures turn out to be common

I expect that this will fix #104.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,6 +66,7 @@ jobs:
     name: Run unit tests
     permissions:
       contents: read
+      id-token: write # to authenticate with codecov
     runs-on: ubuntu-latest
 
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -80,10 +82,18 @@ jobs:
           check-latest: true
       - name: Run Go tests
         run: go test -covermode atomic -coverprofile coverage.txt $(go list ./... | grep -v third_party/)
+      - name: Workaround buggy Codecov OIDC auth
```
cmurphy, line 85: Is there an upstream github issue we can link to so we can remove the workaround when it's fixed?

cmurphy: I didn't think linking it in code would spam them but I could be wrong. I was thinking referring to it in a code comment would help remind us to keep an eye on it, since once #104 is closed we won't think about it again.

cmurphy (approved):

PR #109, opened 2025-03-17
Disable codecov patch updates

haydentherapper:
This creates a decent amount of noisy comments on PRs. We have this off for other repos and codecov still works fine, generating reports for coverage for the PR.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,17 @@
+# Copyright 2025 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+coverage:
+  status:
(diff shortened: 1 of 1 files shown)


jku:
Can you describe what this change does in practice: Does it remove the code annotations or the PR comment, or both?


Examples in sigstore org may not be a good indication of best practices since it seems clear that no one actually looks at codecov coverage (based on the finding that baseline coverage is not being reported in any of the projects).

haydentherapper:
I was under the impression that it would drop the codecov/patch comments left throughout the PR, but still compute overall coverage and provide percent changes, like in https://github.com/sigstore/rekor/pull/2413.

jku (approved):
> I was under the impression that it would drop the codecov/patch comments left throughout the PR, but still compute overall coverage and provide percent changes, like in https://github.com/sigstore/rekor/pull/2413.

This sounds good to me :+1

loosebazooka:
I wonder if we should disable annotations like this instead: https://docs.codecov.com/docs/github-checks#disabling-github-checks-patch-annotations-via-yaml ?

I think patch checks might still be useful, but I don't have a good grasp on how they will be surfaced without annotations.

jku:
Happy to try any setup out -- I don't think I can reliably map the codecov docs to actual GitHub things, they're not to clearest docs ever

loosebazooka:
Also apparently you can just use keyboard shortcut `a` to toggle annotations

cmurphy (approved):

jku:
rebased to get the required checks to appear...

jku:
automerging now: if it does not look like you want after merge, file an issue and I'll investigate

PR #112, opened 2025-03-18
Fix upload success status code

cmurphy:
Use HTTP 201 Created when an entry is successfully added.

### DIFF
--- a/pkg/server/http.go
+++ b/pkg/server/http.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strconv"
 	"sync"
 	"syscall"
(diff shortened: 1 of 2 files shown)


haydentherapper:
Is https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_your_gateway/#:~:text=Additional%20responses%20can%20be%20added%20to%20the%20Protocol%20Buffer%20definitions%20to%20match%20the%20new%20status%20codes%3A something we need to do?

jku (approved):
seems a little weird that grpc-gateway can't manage this but based on the docs you're right

loosebazooka (approved):

PR #113, opened 2025-03-18
Add client methods

cmurphy:
Add a client support package. This will be used in Rekor's integration tests and can be imported by Go-based Sigstore clients.

Depends on #119 

TODO:

- [x] verify inclusion proof from Add response
- [x] add ReadCheckpoint
- [x] add ReadTile
- [x] add ReadEntryBundle
- [x] remove `verifyTLE` and replace with verification functions from https://github.com/sigstore/rekor-tiles/pull/114
- [x] add unit tests

### DIFF
--- a/pkg/client/option.go
+++ b/pkg/client/option.go
@@ -1,0 +1,37 @@
+//
+// Copyright 2025 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+import "time"
+
+type Config struct {
+	UserAgent string
+	Timeout   time.Duration
+}
+
+type Option func(*Config)
+
+func WithUserAgent(agent string) Option {
+	return func(c *Config) {
+		c.UserAgent = agent
+	}
+}
+
+func WithTimeout(timeout time.Duration) Option {
+	return func(c *Config) {
+		c.Timeout = timeout
+	}
+}
--- a/pkg/client/read/read.go
+++ b/pkg/client/read/read.go
@@ -1,0 +1,102 @@
+//
+// Copyright 2025 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package read
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+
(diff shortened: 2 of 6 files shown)


pkg/client/option.go
```diff
@@ -0,0 +1,37 @@
...
+import "time"
+
+type clientConfig struct {
+	userAgent string
+	timeout   time.Duration
+}
+
+type Option func(*clientConfig)
+
+func WithUserAent(agent string) Option {
```
jku, line 27:
```suggestion
func WithUserAgent(agent string) Option {
```

cmurphy:
Will rebase on https://github.com/sigstore/rekor-tiles/pull/114 once merged and replace `verifyTLE` with `verify.VerifyLogEntry`.

jku:
> Will rebase on #114 once merged and replace verifyTLE with verify.VerifyLogEntry.

That is #119 now. I might not have time to get the tests done for that today: if you are in a hurry with this one feel free to use that branch here or merge it, just file an issue so I remember to write those tests...

pkg/client/read.go
```diff
@@ -0,0 +1,102 @@
...
+
+type readClient struct {
+	baseURL  *url.URL
+	client   *tclient.HTTPFetcher
+	origin   string
+	verifier note.Verifier
+}
+
+// NewReader creates a new reader client.
+// (This is currently specific to GCS storage).
```
jku, line 46: why is this?

pkg/client/read.go
```diff
@@ -0,0 +1,102 @@
...
+	baseURL, err := url.Parse(readURL)
+	if err != nil {
+		return nil, fmt.Errorf("parsing url %s: %w", readURL, err)
+	}
+	noteVerifier, err := rekornote.NewNoteVerifier(origin, verifier)
+	if err != nil {
+		return nil, fmt.Errorf("creating note verifier: %w", err)
+	}
+	httpClient := &http.Client{
+		Transport: createRoundTripper(http.DefaultTransport, cfg.userAgent, url.Values{"alt": []string{"media"}}),
```
jku, line 61: what's this query argument? is this the GCS specific bit?

jku (commented):
Looks great to me. 

Are you planning to add tests for the other read client functions or no?

pkg/client/read.go
cmurphy: Yes, without this the request just returns the object metadata instead of the object. I'll add a comment.

cmurphy:
Added tests for the other read methods (should have done that before) and added comments.

pkg/client/read.go
jku: I see. Is this handled by the load balancer somehow in the actual deployment?

cmurphy: Options for an actual deployment are pointing the read client directly to the GCS bucket URL or to a load balancer URL rewriter, but either way the query parameters will need to be passed through to get the raw object.

jku: requirement for sigstore clients to inject a GCS specific query parameter seems quite unappealing, no? 

I suppose we can setup a `google_compute_backend_bucket` on the load balancer like https://tuf-repo-cdn.sigstore.dev/ does: something there has to manage this bit since it just works.

cmurphy: Thanks for noticing this, there's another endpoint we can use that doesn't need this query parameter https://cloud.google.com/storage/docs/request-endpoints#access_to_public_objects will change it to use that

jku: Actually, I think you are mistaken: this works:
https://storage.googleapis.com/tuf-root-staging/11.root.json

(thats https://storage.googleapis.com/<bucket>/<object>)

cmurphy: Fixed

loosebazooka (dismissed):
Would it make sense to separate out the clients so the import trees are smaller for write and read clients? Is that even a problem in go?

cmurphy:
> Would it make sense to separate out the clients so the import trees are smaller for write and read clients? Is that even a problem in go?

Done. That is a thing in Go but it didn't seem like that much overhead in this case. But that could change so I moved them.

jku (approved):
lgtm

PR #116, opened 2025-03-20
Make SafeInt64 public

cmurphy:
Expose the uint64<->int64 conversion type so it can be used more widely over the code base.

### DIFF
--- a/pkg/tessera/safeint.go
+++ b/pkg/tessera/safeint.go
@@ -20,13 +20,13 @@ import (
 	"math"
 )
 
-// safeInt64 holds equivalent int64 and uint64 integers.
-type safeInt64 struct {
+// SafeInt64 holds equivalent int64 and uint64 integers.
+type SafeInt64 struct {
 	u uint64
 	i int64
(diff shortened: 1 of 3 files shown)


pkg/tessera/safeint.go
```diff
@@ -20,13 +20,13 @@ import (
 	"math"
 )
 
-// safeInt64 holds equivalent int64 and uint64 integers.
-type safeInt64 struct {
-	u uint64
-	i int64
+// SafeInt64 holds equivalent int64 and uint64 integers.
+type SafeInt64 struct {
+	U uint64
```
loosebazooka, line 25: should these still be un-exported with getters (`U()`, `I()`)? I worry someone would just set them and pass SafeInt64 around externally, without using NewSafeInt64?

loosebazooka (dismissed):
I understand if that sounds like overkill coming from a java person though

cmurphy:
No I get it, fixed

loosebazooka (approved):

PR #117, opened 2025-03-20
Add metrics

loosebazooka:
Add default metrics to grpc and http servers
Add some rudimentary metrics for counting new entries

Serve metrics on (default) 0.0.0.0:2112 or <host>:<metrics-port>

#### Summary
Adds prometheus config

Endpoints using configs should add new metric types to metrics, and increment them (or whatever) by using the singleton instance provided by getMetrics()

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/pkg/server/metrics.go
+++ b/pkg/server/metrics.go
@@ -1,0 +1,162 @@
+// Copyright 2025 The Sigstore Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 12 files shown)


pkg/server/metrics.go
```diff
@@ -0,0 +1,162 @@
...
+			ReadTimeout:       60 * time.Second,
+			ReadHeaderTimeout: 60 * time.Second,
+			WriteTimeout:      60 * time.Second,
+			IdleTimeout:       config.idleTimeout,
+		},
+		serverEndpoint: endpoint,
+	}
+}
+
+func (hp *httpMetrics) start(wg *sync.WaitGroup) {
```
loosebazooka, line 125: This is very much like all the other servers, although I was looking at `github.com/oklog/run` to simplify our signal handling and server shutdowns (I dunno).

jku:
If you want to test a tweak to the codecov auth workaround: https://github.com/jku/rekor-tiles/commit/88488b4278e1a2d1bab02d6ae20e27b827d5cdb1
* this seems like a codecov issue to me: they just see branch on origin repo and assume they need a token even though they don't since we just want the PR results that work fine without a token
* I pretty much never use PRs from same repo but I think they do have access to the oidc token... I have also not tested this since I can't make a branch but vibes are that it might work

loosebazooka:
switching to a fork

PR #122, opened 2025-03-21
workflows: Tweak oidc auth for codecov once more

jku:
Currently codecov step succeeds in main and in PRs from forks, but fails in PRs from origin. This seems to be a codecov bug but I believe we can tweak the workaround a bit.

* Always run the auth workaround step
* Only try to set the codecov token if OIDC is available

This should mean that:
* PRs from forks will not have auth token: that works fine
* pushes to main will have auth token
* PRs from origin repo will have auth token (even though they should not need it)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,8 +83,10 @@ jobs:
       - name: Run Go tests
         run: go test -covermode atomic -coverprofile coverage.txt $(go list ./... | grep -v third_party/)
       - name: Workaround buggy Codecov OIDC auth
-        if: github.event_name == 'push'
         run: |
+          # only set CODECOV_TOKEN if OIDC token is available
+          [ -z $ACTIONS_ID_TOKEN_REQUEST_TOKEN ] && exit 0
+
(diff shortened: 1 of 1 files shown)


jku:
This works in a PR within my fork. If it works in this PR too, both cases are covered.

.github/workflows/test.yml
```diff
@@ -83,8 +83,10 @@ jobs:
       - name: Run Go tests
         run: go test -covermode atomic -coverprofile coverage.txt $(go list ./... | grep -v third_party/)
       - name: Workaround buggy Codecov OIDC auth
-        if: github.event_name == 'push'
         run: |
+          # only set CODECOV_TOKEN if OIDC token is available
+          [ -z $ACTIONS_ID_TOKEN_REQUEST_TOKEN ] && exit 0
```
loosebazooka, line 88: could we do `if: ${{ env.ACTIONS_ID_TOKEN_REQUEST_TOKEN != '' }}`

loosebazooka (approved):
Seems fine, just a minor curiosity

.github/workflows/test.yml
jku: I can try it out, I guess it's a little more visible to users

jku: actually, it doesn't work: I remember now that environment variables are not available in the task conditional.

loosebazooka (approved):

Issue #10, opened 2025-02-24
Add entry verifiers (copy from Rekor v1)


haydentherapper:
**Description**

cmurphy:
This isn't actually needed, we're not signing entries in rekor v2, only checkpoints.

haydentherapper:
@cmurphy I thought this was referring to copying over some interfaces and implementations from https://github.com/sigstore/rekor/tree/main/pkg/pki, which is needed to verify signatures on entries.

If so, I'd suggest we only copy over `x509`, and I'd recommend we split it into two verifiers, certificate verification and public key verification. We can add other signature verifiers as we see fit.

PR #144, opened 2025-03-25
Rework proto defs, add log types

### LINKED ISSUE #145: Create log types that correspond to requests
Add corresponding log entry types for each request type
Also version the types

ex: HashedRekordRequestV0_0_2 -> HashedRekordLogentryV0_0_2 ...

loosebazooka:
- Add version suffixes to all types
- Make dsseEnvelope strongly typed at request time
- Add proto defs for types that are persisted to the log

#### Summary
I thought this would be simply adding versions to types, but I found a few more things to type a little more strictly

fixes #145 

#### Release Note
NONE

#### Documentation
NONE


api/proto/dsse.proto
```diff
@@ -16,16 +16,25 @@ syntax = "proto3";
...
-    repeated Verifier verifier = 2;
+    repeated Verifier verifiers = 2 [(google.api.field_behavior) = REQUIRED];
 }
+
+
+message DSSELogEntryV0_0_2 {
+    // The hash of the dsse payload
+    dev.sigstore.common.v1.HashOutput payloadHash = 2 [(google.api.field_behavior) = REQUIRED];
+    // A collection of all signatures over the payload
+    repeated bytes signatures = 3 [(google.api.field_behavior) = REQUIRED];
```
cmurphy, line 39: I think this needs the verifier too. The old DSSE had that in the persisted object https://github.com/sigstore/rekor/blob/73dba7c07d0747f00119417fc0ff994a393f97b2/pkg/types/dsse/v0.0.1/entry.go#L235-L238

pkg/types/dsse/dsse.go
```diff
@@ -15,23 +15,49 @@
...
 		}
 	}
-	return nil
+
+	if ds.Envelope.Payload == nil {
+		return nil, fmt.Errorf("missing envelope payload")
+	}
+	payloadHash := sha256.Sum256([]byte(ds.Envelope.Payload))
+
+	// TODO: check that signatures can be validated against the provided keys
```
cmurphy, line 44: I'm working on this in https://github.com/sigstore/rekor-tiles/issues/10 in case you haven't seen it

loosebazooka: ah okay cool. I'll mark the TODO with `#10` on it.

api/proto/dsse.proto
```diff
@@ -16,16 +16,26 @@ syntax = "proto3";
...
 // A request to add a dsse entry to the log
-message DSSERequest {
-    // A DSSE envelope (as defined by https://github.com/secure-systems-lab/dsse/blob/master/envelope.proto)
-    // in stringified JSON
-    string envelope = 1 [(google.api.field_behavior) = REQUIRED];
-    // collection of verification material used to verify
-    repeated Verifier verifier = 2;
+message DSSERequestV0_0_2 {
+    // A DSSE envelope
+    io.intoto.Envelope envelope = 1 [(google.api.field_behavior) = REQUIRED];
```
haydentherapper, line 29: [This envelope](https://github.com/sigstore/protobuf-specs/blob/7e4a1adcf03949f345a9f37c62de8a8df8dac128/protos/envelope.proto#L41) includes signatures, so would you want only the payload and payload type?

haydentherapper (commented):
What is the goal of requiring the client to include the mapping between verifier and signature for DSSE, to reduce server complexity?

We currently mandate that the bundle include only one signature for a DSSE - https://github.com/sigstore/protobuf-specs/blob/7e4a1adcf03949f345a9f37c62de8a8df8dac128/protos/sigstore_bundle.proto#L143. I don't think we should enforce the same server-side, to handle private deployments, but then this would require custom clients that can build the verifier-sig mapping, rather than just let the server figure it out.

loosebazooka:
> What is the goal of requiring the client to include the mapping between verifier and signature for DSSE, to reduce server complexity?

Yeah this was actually a bit murky, but it was to ensure that the mapping was provided before doing the weird custom logic in the server to map keys to signatures. It provided some uniformity in the api for signature though.

On thursday we discussed this, I'm basically okay either way, but having non-sigstore clients doing a bunch of work to write dsse to rekor is a valid point (however, it's not that much extra work to do this, they have be aware of the request format anyway). The only thing I see here is if their signer is separate from the rekor client, otherwise they've already parsed the envelope.

loosebazooka (commented):
Rebased and updated tests to reflect log types returned.

api/proto/dsse.proto
```diff
@@ -16,16 +16,26 @@ syntax = "proto3";
...
+    // Signatures and their associated verification material used to verify the payload in the envelope
+    // These should be a direct 1:1 mapping with the signatures provided in the envelope
```
haydentherapper, line 31: Would we err out if there's an unequal number of verifiers and signatures? If not, I might update wording to say that all signatures must be verified using verifiers (so that additional verifiers could be specified and that's not an error)

api/proto/dsse.proto
haydentherapper: LGTM, at the time I was concerned that we would duplicate signatures with a `SignatureAndVerifier` in the request

api/proto/dsse.proto
```diff
@@ -16,16 +16,26 @@ syntax = "proto3";
...
+    repeated Verifier verifiers = 2 [(google.api.field_behavior) = REQUIRED];
+}
+
+
+message DSSELogEntryV0_0_2 {
+    // The hash of the dsse payload
```
haydentherapper, line 37:
```suggestion
    // The hash of the DSSE payload
```

pkg/types/dsse/dsse.go
```diff
@@ -18,67 +18,76 @@ import (
...
 	"github.com/sigstore/rekor-tiles/pkg/pki/x509"
 	"github.com/sigstore/rekor-tiles/pkg/types/verifier"
 	"github.com/sigstore/sigstore/pkg/signature"
 	sigdsse "github.com/sigstore/sigstore/pkg/signature/dsse"
 )
 
-func Validate(ds *pb.DSSERequest) error {
-	if ds.Envelope == "" {
-		return fmt.Errorf("missing envelope")
+func ToLogEntry(ds *pb.DSSERequestV0_0_2) (*pb.DSSELogEntryV0_0_2, error) {
```
haydentherapper, line 37: I think this should be split into two functions, one for proto validity, and one for cryptographic validity, or we need a name that conveys that more clearly.

pkg/types/dsse/dsse.go
```diff
@@ -18,67 +18,76 @@ import (
...
+		},
+		Signatures: signerVerifiers,
+	}, nil
 }
 
 // verifyEnvelope takes in an array of possible key bytes and attempts to parse them as x509 public keys.
 // it then uses these to verify the envelope and makes sure that every signature on the envelope is verified.
-// it returns a map of verifiers indexed by the signature the verifier corresponds to.
+// it returns a list of SignatureAndVerifier mapping each signature in the envelope to a provided verifier
 // Copied from https://github.com/sigstore/rekor/blob/73dba7c07d0747f00119417fc0ff994a393f97b2/pkg/types/dsse/v0.0.1/entry.go#L364-L403
```
haydentherapper, line 80:
```suggestion
// Copied and modified from https://github.com/sigstore/rekor/blob/73dba7c07d0747f00119417fc0ff994a393f97b2/pkg/types/dsse/v0.0.1/entry.go#L364-L403
```

api/proto/hashedrekord.proto
```diff
@@ -23,11 +23,16 @@ import "verifier.proto";
...
-    // A single signature over the hashed data
-    bytes signature = 1 [(google.api.field_behavior) = REQUIRED];
+message HashedRekordRequestV0_0_2 {
     // The hashed data
-    dev.sigstore.common.v1.HashOutput data = 2 [(google.api.field_behavior) = REQUIRED];
-    // A key/cerificate to verify the signature
-    Verifier verifier = 3 [(google.api.field_behavior) = REQUIRED];
+    dev.sigstore.common.v1.HashOutput data = 1 [(google.api.field_behavior) = REQUIRED];
+    // A single signature over the hashed data with the verifier needed to validate it
+    SignatureAndVerifier signature = 2 [(google.api.field_behavior) = REQUIRED];
```
haydentherapper, line 30: Do we want to rename `signature` since it also includes a verifier, here and in the entry?

haydentherapper (commented):
just small comments

pkg/types/dsse/dsse_test.go
```diff
@@ -137,7 +204,7 @@ func TestValidate(t *testing.T) {
 	}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			gotErr := Validate(test.dsse)
+			_, gotErr := ToLogEntry(test.dsse)
```
cmurphy, line 207: Do you want to check the converted value here?

docs/openapi/rekor_service.swagger.json
```diff
@@ -512,61 +549,73 @@
...
       "title": "Create a new HashedRekord or DSSE",
       "required": [
-        "hashedRekordRequest",
-        "dsseRequest"
+        "hashedRekordRequestV002",
+        "dsseRequestV002"
       ]
     },
-    "v2DSSERequest": {
+    "v2DSSERequestV0_0_2": {
```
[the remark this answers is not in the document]
loosebazooka: That v2 prefix is the package name, the v2_0_0 is the version of the request type. I guess it reads weird, but its a reference to a type definition within the swagger docs, the package prefix (v2) does not appear in the json. We see the same thing in the [fulcio swagger output](https://github.com/sigstore/fulcio/blob/main/fulcio.swagger.json#L121).

api/proto/hashedrekord.proto
loosebazooka: yeah maybe, our v1 api used
```
signature: {
   content:...
   public_key: {
      content:...
   }
}
```

but yeah we could rename this to `signatureAndVerifier`

loosebazooka: actually I think we can just structure this like
```
signature: {
   content:...
   verifier: {
      content:...
   }
}
```

haydentherapper: I’d be good with that too, I like that over a longer name.

pkg/types/dsse/dsse.go
loosebazooka: Sure? There's a lot of work in validate that is useful to conversion. I could just make the conversion function always call validate before. But logically it seems like we'd be unlikely to call validate and tologentry separately

haydentherapper: I’m not super opinionated on if we should refactor into smaller functions, but this name isn’t conveying that this function does a) validation of request input, b) signature verification, c)converting request to entry. I think we could have each of those as unexported functions (which we can export later if they’d be useful in a library) and then one public function that calls all three.

haydentherapper: I also like this design because we can decide later to add back something like the Entry interface (which is overkill now since we have just two types)

loosebazooka: updated with comment, will followup in #178

cmurphy (approved):

PR #149, opened 2025-03-27
use go tool to manage versions of tools

bobcallaway:


loosebazooka:
There was some discussion about not polluting the dependency versions with gotool, but I don't have a strong opinion here.

cmurphy (dismissed):

Issue #148, opened 2025-03-26
Expose batching parameters


cmurphy:
**Description**

Add flags in the Rekor server to change these defaults in Tessera:

https://github.com/transparency-dev/trillian-tessera/blob/a60c8d23e801e11eb7511a623d144d6a7da3d0d4/append_lifecycle.go#L35-L42

PR #152, opened 2025-03-27
Expose Tessera append lifecycle options

cmurphy:
Expose Tessera's batch size, batch age, checkpoint interval, and pushback options through Rekor's configuration. This will enable us to tune the server to balance the cost and latency of sequencing entries, and in particular it will allow us to reduce the time to publish a new checkpoint so that entries added sequentially in tests no longer need to wait up to 10 seconds for a response.

Fixes https://github.com/sigstore/rekor-tiles/issues/148

### DIFF
--- a/cmd/rekor-server/app/serve.go
+++ b/cmd/rekor-server/app/serve.go
@@ -72,7 +72,12 @@ var serveCmd = &cobra.Command{
 			slog.Error(fmt.Sprintf("failed to initialize signer: %v", err.Error()))
 			os.Exit(1)
 		}
-		tesseraStorage, err := tessera.NewStorage(ctx, viper.GetString("hostname"), tesseraDriver, signer)
+		appendOptions, err := tessera.NewAppendOptions(ctx, viper.GetString("hostname"), signer, viper.GetUint("batch-max-size"), viper.GetDuration("batch-max-age"), viper.GetDuration("checkpoint-interval"), viper.GetUint("pushback-max-outstanding"))
+		if err != nil {
+			slog.Error(fmt.Sprintf("failed to initialize append options: %v", err))
+			os.Exit(1)
+		}
+		tesseraStorage, err := tessera.NewStorage(ctx, viper.GetString("hostname"), tesseraDriver, appendOptions)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/rekor-server/app/serve.go
```diff
@@ -116,6 +121,13 @@ func init() {
 	serveCmd.Flags().String("signer-kmshash", "sha256", "hash algorithm used by the KMS")
 	serveCmd.Flags().String("signer-tink-kek-uri", "", "encryption key for decrypting Tink keyset. Valid options are [aws-kms://keyname, gcp-kms://keyname]")
 	serveCmd.Flags().String("signer-tink-keyset-path", "", "path to encrypted Tink keyset")
+
+	// tessera lifecycle configs
+	serveCmd.Flags().Uint("batch-max-size", tessera.DefaultBatchMaxSize, "the maximum number of entries that will accumulated before being sent to the sequencer")
+	serveCmd.Flags().Duration("batch-max-age", tessera.DefaultBatchMaxAge, "the maximum amount of time a batch of entries will wait before being sent to the sequencer")
+	serveCmd.Flags().Duration("checkpoint-interval", tessera.DefaultCheckpointInterval, "the frequency at which a checkpoint will be published")
```
jku, line 128: Is it somehow visible in CLI that these are duration strings that get parsed?

jku (approved):
lgtm with one question on CLI discoverability

cmd/rekor-server/app/serve.go
bobcallaway:
```sh
start the Rekor server

Usage:
  rekor-server serve [flags]

Flags:
      --batch-max-age duration           the maximum amount of time a batch of entries will wait before being sent to the sequencer (default 250ms)
      --batch-max-size uint              the maximum number of entries that will accumulated before being sent to the sequencer (default 256)
      --checkpoint-interval duration     the frequency at which a checkpoint will be published (default 10s)
```

jku: with the example in the default value that seems clear enough

ramonpetgrave64 (approved):

PR #153, opened 2025-03-27
feat:  add timeouts, max payload size

ramonpetgrave64:
#### Summary

 - Adds timeouts for the GRPC server.
 - Adds max payload size, and metrics for the payload sizes

The timeout can be configured at CLI with `--grpc-timeout=30s`, for example.

We use this value for both when establishing a connection and when the connection is established but idle.

 - https://grpc.io/docs/guides/keepalive/#keepalive-configuration-specification

**Testing Process**

No new unit tests added against these changes, specifically. It is difficult to test the timeouts, but we can manually test the max payload size.

```shell
payload_size=$((5 * 1024 * 1024)) 
payload_file="payload.json"
random_data=$(head -c "$payload_size" /dev/urandom | base64)
json_payload=$(cat <<EOF
{
  "data": "$random_data"
}
EOF
)
echo "$json_payload" > "$payload_file"
curl -X POST \
  -H "Content-Type: application/json" \
  -d @- \
  http://localhost:3000/api/v2/log/entries < payload.json 

{"code":3, "message":"http: request body too large", "details":[]}%   
```

**TODO**:

- [x] since the grpc server is fronted by an http proxy , determine if it makes sense to set the grpc server's connection and idle timeouts.
- [x] determine a good [MaxRecvMsgSize](https://pkg.go.dev/google.golang.org/grpc@v1.71.0#MaxRecvMsgSize)
- [x] A way to test the timeouts and payload sizes

#### Release Note

Adds timeouts, max payload size, and payload size metrics.


haydentherapper:
Discussion from 4/1 meeting: HTTP proxies to gRPC, so we should test what the error is when gRPC times out for an HTTP request. We can also configure the idle timeout for HTTP to match the gRPC timeout [here](https://github.com/sigstore/rekor-tiles/blob/258d6f19133fb45de2187557f8b8800ad6f635ad/pkg/server/http.go#L102), which is what we do in Fulcio.

re: max receiving size, gRPC defaults to 4Mb. We have a hard limit of 128Mb at the load balancer - https://github.com/sigstore/public-good-instance/blob/33d9d6f53441ead73b2ba60dfa15c5d2329634b6/terraform/environments/staging/1-infrastructure/ingress.tf#L229

pkg/server/grpc.go
```diff
@@ -51,9 +52,13 @@ type grpcServer struct {
...
 
-	opts = append(opts, grpc.ChainUnaryInterceptor(
-		getMetrics().serverMetrics.UnaryServerInterceptor(),
-	))
+	opts = append(opts,
+		grpc.ChainUnaryInterceptor(getMetrics().serverMetrics.UnaryServerInterceptor()),
+		grpc.ConnectionTimeout(config.timeout),
+		grpc.KeepaliveParams(keepalive.ServerParameters{MaxConnectionIdle: config.timeout}),
+		// explicitly set to 4MB. https://github.com/grpc/grpc-go/blob/cdbdb759dd67c89544f9081f854c284493b5461c/server.go#L59C39-L59C54
+		grpc.MaxRecvMsgSize(defaultMaxSizeBytes),
```
haydentherapper, line 59: can this be a flag value, `max-request-body-size`, and passed through configuration?

ramonpetgrave64: There is also the DefaultMaxHeaderBytes, which by default is 1MB, but some providers like GCP load balancer limit it to 8KB. Should we also make that configurable? 

- https://pkg.go.dev/net/http#pkg-constants
- https://cloud.google.com/load-balancing/docs/https/custom-headers#:~:text=the%20same%20name.-,Limitations,KB%20or%2016%20request%20headers.

haydentherapper: That default seems fine and not something we'd need to modify.

haydentherapper (approved):

PR #155, opened 2025-03-28
build: Add vscode workspace settings

jku:
Without this the language server can't handle tests/e2e_tests.go from #128

(if everyone else is some kind of emacs wizard I can keep this to myself too)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "go.buildTags": "e2e",
+}


cmurphy:
Neither fulcio nor rekor v1 have this, how do vscode users manage for those projects?

jku:
this is (almost) my first golang project so not sure but I  assume they add required build tags to their personal settings (or live without ide support in tests)

haydentherapper:
I usually just run `go test -tags=e2e ./...` from the command line. I don't have a custom vscode settings file so I wouldn't be opposed to this, but this could cause a conflict for another developer who does have a custom file.

Just searching around for guidance, https://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control suggests checking this file in.

haydentherapper (approved):

jku:
> I usually just run go test -tags=e2e ./... from the command line. I don't have a custom vscode settings file so I wouldn't be opposed to this, but this could cause a conflict for another developer who does have a custom file.

I was unclear: the issue is that basic annotations and autocomplete in the IDE are missing in `e2e_tests.go` without this setting -- so either we commit this or all vscode users do a similar change in their user settings.

I'll merge this now, let's revisit if anyone complains.
.

Issue #157, opened 2025-03-28
Add persistent deduplication


cmurphy:
**Description**

Add support for [`WithAntispam`](https://github.com/transparency-dev/trillian-tessera/blob/61661a3e39541adec35d4eaaf02c983f2fff0e60/append_lifecycle.go#L203) which uses a separate Spanner table to prevent duplicated entries. This brings us into feature parity with Rekor v1/Trillian which did not allow duplicate entries.

PR #159, opened 2025-03-28
Add support for deduplication

cmurphy:
Add options to configure the WithAntispam Tessera append options[1].

The "persistent" antispam layer does not work with the GCP Spanner emulator because it relies on the currently unsupported "batch write" operation. So this option is disabled by default and not tested in the e2e tests, but should be enabled when deployed to real infrastructure.

[1] https://github.com/transparency-dev/trillian-tessera?tab=readme-ov-file#antispam

This updates Tessera to an unreleased version in order to get the latest antispam settings.

Fixes https://github.com/sigstore/rekor-tiles/issues/157


haydentherapper (approved):

cmurphy:
> If the same request hits our loud balancer and is farmed out to two rekor-tiles instances, do we end up with duplicate entries?

Yes, this level would not protect against that. The "persistent" layer should.

haydentherapper (approved):

PR #161, opened 2025-03-31
Shutdown Tessera gracefully on SIGTERM/SIGINT

### LINKED ISSUE #138: Ensure SIGTERM is caught and graceful shutdown is appropriately handled
**Description**

In K8S, pods are sent `SIGTERM` when they're being stopped by the scheduler (after new requests have been routed away from the pod), and then waits `terminationGracePeriodSeconds` before sending `SIGKILL`; we need to determine how to appropriately handle this situation. For example:

- Ensure the sequencer job runs at least once?
- Then?, ensure a checkpoint is published?
- Clean up database connections, GCS client?

haydentherapper:
Fixes #138

When SIGTERM/SIGINT is sent to the process, which occurs when Kubernetes is shutting down pods, the server gracefully shuts down the gRPC and HTTP servers. Tessera also provides a shutdown function which will wait until a checkpoint has been published that covers all sequenced entries.

This change plumbs this shutdown function through to call it when the server is shutting down.

### DIFF
--- a/pkg/server/serve.go
+++ b/pkg/server/serve.go
@@ -45,7 +45,9 @@ func Serve(ctx context.Context, hc *HTTPConfig, gc *GRPCConfig, s rekorServer, t
 
 	wg.Wait()
 
+	slog.Info("shutting down Tessera sequencer")
 	if err := tesseraShutdownFn(ctx); err != nil {
 		slog.Error("error shutting down Tessera", "error", err)
 	}
+	slog.Info("stopped Tessera sequencer")
 }
--- a/pkg/server/serve.go
+++ b/pkg/server/serve.go
@@ -22,7 +22,7 @@ import (
 )
 
 // Serve starts the grpc server and its http proxy.
-func Serve(ctx context.Context, hc *HTTPConfig, gc *GRPCConfig, s rekorServer) {
+func Serve(ctx context.Context, hc *HTTPConfig, gc *GRPCConfig, s rekorServer, tesseraShutdownFn func(context.Context) error) {
 	var wg sync.WaitGroup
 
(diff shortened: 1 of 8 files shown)


haydentherapper:
Also disabling codecov project coverage. It still shows it in the comment, but won't cause a GitHub Actions failure if there's a percent drop, which this PR had a 0.06% drop. We did this in other Sigstore repos too.

haydentherapper:
Note that arguably this does little at the moment, as the HTTP and gRPC shutdown methods should wait until a checkpoint past the largest log index is published since we use `Await`, and the shutdown methods won't terminate the server until all connections are closed. This would be more meaningful for a log that doesn't use Await.

We could invert the order and call the Tessera shutdown method before shutting down the HTTP and gRPC connections, but this only blocks new writes and not reads, and if the Tessera shutdown adds support for closing Spanner and GCS connections as I requested, then this would cause errors for new read requests.

pkg/server/serve.go
```diff
@@ -44,4 +44,8 @@ func Serve(ctx context.Context, hc *HTTPConfig, gc *GRPCConfig, s rekorServer) {
 	httpMetrics.start(&wg)
 
 	wg.Wait()
+
+	if err := tesseraShutdownFn(ctx); err != nil {
```
jku, line 48: Do we end up with good enough logging that we will be able to tell that we are now waiting for tessera to shutdown, or do we need to explicitly log that?

cmurphy (commented):
lgtm apart from Jussi's question

haydentherapper:
Added log messages.

cmurphy (dismissed):

jku (approved):

PR #163, opened 2025-03-31
Fix metrics test failure when run by itself

### LINKED ISSUE #162: metrics_test fails when run by itself
**Description**

Reproduced with `go test -timeout 30s -tags e2e -run ^TestServe_httpMetricsSmoke$ github.com/sigstore/rekor-tiles/pkg/server`

The test fails because `rekor_http_api_latency` and `rekor_http_requests_total` are not found in the metrics body output.

When running all tests at once with `go test ./...`, this test succeeds.

I'm assuming this failure is because another test will set these metrics.

**Version**

https://github.com/sigstore/rekor-tiles/commit/c7178673409d48b58fef58466d5391dd2017515b

haydentherapper:
Fixes #162

This failure occurs because the HTTP latency and request metrics seem to be created only when a request has occurred.

When running all tests with go test ./..., no failure occurs, which is why CI didn't catch this. I believe this succeeded because another test triggered an HTTP request and however Prometheus is creating and scraping metrics created shared state between tests.

### DIFF
--- a/pkg/server/metrics_test.go
+++ b/pkg/server/metrics_test.go
@@ -29,12 +29,18 @@ func TestServe_httpMetricsSmoke(t *testing.T) {
 	server.Start(t)
 	defer server.Stop(t)
 
+	// Call an endpoint to create HTTP latency and request metrics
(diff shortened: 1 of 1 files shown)


jku (approved):
makes sense

Issue #14, opened 2025-02-24
Add configure options and commands to help support the sharding procedure


haydentherapper:
**Description**

See Step 6 from the [design doc](https://docs.google.com/document/d/1qZ-lkpbQkBzV45rtemWYmT6ReqCwjTt5TbMDFLdaPyM/edit?resourcekey=0-bMAyN9EKPDvB0H3edYi_Cw&tab=t.0#heading=h.p7n5pw5vt1kw).

PR #168, opened 2025-04-01
Add support for freezing the log

cmurphy:
Add a flag that, when Rekor is restarted, prevents the service from accepting new entries or publishing new checkpoints. Add a command to publish a final checkpoint that contains an extension indicating the log is frozen.

Technically Rekor could be fully shut down and it would accomplish the same purpose of not accepting new entries or updating new checkpoints. The server change is mostly useful in the potential future when Rekor needs to be responsible for serving checkpoints stored in MySQL. The freeze command is necessary no matter the backend.

Fixes https://github.com/sigstore/rekor-tiles/issues/14

To do:
- [x] needs a verifier loading flow similar to everything in pkg/signer that supports keys on disk, keys with passwords, KMSs and Tink.
- [x] end to end tests
- [ ] file issue to document the frozen string

### DIFF
--- a/cmd/freeze-checkpoint/app/root.go
+++ b/cmd/freeze-checkpoint/app/root.go
@@ -1,0 +1,232 @@
+//
+// Copyright 2025 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"bytes"
+	"context"
+	"crypto"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"strings"
+	"time"
+
+	gcs "cloud.google.com/go/storage"
+	rekornote "github.com/sigstore/rekor-tiles/pkg/note"
+	"github.com/sigstore/rekor-tiles/pkg/signerverifier"
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	logformat "github.com/transparency-dev/formats/log"
+	"github.com/transparency-dev/trillian-tessera/api/layout"
+	"golang.org/x/mod/sumdb/note"
+)
+
+const (
+	frozenString = "Log frozen — "
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "freeze-checkpoint",
+	Short: "Freeze the log checkpoint",
+	Long:  `Add an extension line to the final checkpoint to indicate to consumers that no more checkpoints are going to be published. Only supported for GCP.`,
+	Run: func(cmd *cobra.Command, _ []string) {
+		ctx := cmd.Context()
+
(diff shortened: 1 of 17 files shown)


pkg/server/http.go
```diff
@@ -184,16 +190,43 @@ func httpResponseModifier(ctx context.Context, w http.ResponseWriter, _ proto.Me
...
+		_, err := w.Write([]byte(msg + "\n"))
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
+
+func methodNotAllowedHandler(ctx context.Context, mux *runtime.ServeMux, marshaler runtime.Marshaler, w http.ResponseWriter, r *http.Request, err error) {
```
haydentherapper, line 218: Can you add a comment for why we need this, that we're remapping a gRPC code to a different HTTP code?

cmd/freeze-checkpoint/app/root.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, fmt.Errorf("parsing checkpoint: %w", err)
+	}
+	if strings.Contains(string(rest), frozenString) {
+		return nil, nil
+	}
+	return &checkpoint, nil
+}
+
+func updateCheckpoint(objWriter *gcs.Writer, noteSigner note.Signer, checkpoint *logformat.Checkpoint) error {
+	buf := bytes.NewBuffer(checkpoint.Marshal())
```
haydentherapper, line 217: tiny nit, can you add a comment that this marshalled checkpoint is just the note and not the signatures? I double checked, wanted to make sure we aren't appending after the sigs

haydentherapper (commented):
This is great!

cmd/freeze-checkpoint/app/root.go
```diff
@@ -0,0 +1,234 @@
...
+		return nil, nil, fmt.Errorf("getting GCS client: %w", err)
+	}
+	bucketName := viper.GetString("gcp-bucket")
+	object := client.Bucket(bucketName).Object(layout.CheckpointPath)
+	objReader, err := object.NewReader(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("getting object reader: %w", err)
+	}
+	contentType := "text/plain; charset=utf-8"
+	cacheControl := "no-cache"
```
haydentherapper, line 189: Could we cache it indefinitely since it should never be updated again?

haydentherapper (approved):

haydentherapper:
I know you mentioned it, we'll want to track this along with the change in key IDs for non-ECDSA signed checkpoints and additional signatures for witnesses for client conformance testing.

haydentherapper (approved):
🥳

haydentherapper:
image

i don't know, i'm bad at memes

cmd/freeze-checkpoint/app/root.go
```diff
@@ -0,0 +1,232 @@
...
+	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	logformat "github.com/transparency-dev/formats/log"
+	"github.com/transparency-dev/trillian-tessera/api/layout"
+	"golang.org/x/mod/sumdb/note"
+)
+
+const (
+	frozenString = "Log frozen — "
```
cmurphy, line 41: @haydentherapper before I merge this do you have any thoughts on using this string + date to indicate the frozen log? A monitor is going to have to be checking for this string so human readability isn't the only consideration.

haydentherapper: This looks good, monitors can easily check for this constant string. I'm not sure if the timestamp will be necessary since we will be co-signing the checkpoint and getting signed timestamps later on, but it's helpful now and would be fine in the future too.

PR #177, opened 2025-04-03
Delete remaining gRPC headers from HTTP response

### LINKED ISSUE #97: Strip grpc headers from http and set response codes
**Description**

cmurphy:
The Grpc-Metadata-Content-Type header was still present in the HTTP response, this removes it.

Fixes #97

### DIFF
--- a/pkg/server/http.go
+++ b/pkg/server/http.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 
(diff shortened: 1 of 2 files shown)


pkg/server/http.go
```diff
@@ -190,6 +191,12 @@ func httpResponseModifier(ctx context.Context, w http.ResponseWriter, _ proto.Me
 		return nil
 	}
 
+	for header := range w.Header() {
+		if strings.HasPrefix(header, "Grpc-") {
+			delete(w.Header(), header)
```
haydentherapper, line 196: Do we need to remove `		delete(w.Header(), "Grpc-Metadata-X-Http-Code")` below?

haydentherapper (approved):

Issue #176, opened 2025-04-03
Allow dsse request to include requested hash algorithm


loosebazooka:

loosebazooka:
Instead of getting the hash algorithm from the user, we should just infer from the signature scheme what it is and hash the payload using the same hash.

@cmurphy this probably means we need the algorithm registry ready and some sort of key hashing algorithm parser ready to go.

haydentherapper:
To add onto this, we will need to handle inferring the digest alg from the signature scheme for both keys and certificates.

For keys, we can use the algorithm registry to look up hash by `PublicKeyDetails`.

For certificates, we will need to add OIDs to the registry mapping. For hashedrekord, we will need to special case the lookup for ed25519-ph. There's no OID for ed25519-ph, so Fulcio certificates will contain the OID for ed25519. So we just need a special case for "request is hashedrekord and OID is ed25519, then use ed25519-ph".

Also, for hashedrekord, we should change the request. Digest input should not be a `HashOutput`, instead just the digest bytes. Otherwise, we have two places where the hash alg is specified, but will be ignoring one of them.

haydentherapper:
Spent some time today on this and realized certificates don't give us what we need. A certificate's SignatureAlgorithm is how the certificate is signed. The certificate contains the subject public key info, but this is just the key algorithm, nothing about the digest algorithm that the caller will use when signing with the key.

For DSSE, we need to specify the hash algorithm the user used to sign the DSSE, which will be used to compute the payload hash both for signature verification and storing in the log.

For when a key is specified, I think we actually should not use the `PublicKey` structure, and just take in `bytes` as a DER-encoded public key. We don't use `PublicKey.ValidFor`, and we can derive the `PublicKeyDetails` through a) the key algorithm, b) the key size, c) the provided hash from `HashOutput` in the request, and for the special case of ed25519 vs ed25519-ph, d) if the request is for a hashedrekord or not.

PR #179, opened 2025-04-03
Add dsse hashing option to request

loosebazooka:
#### Summary
Allow the requester to chose how to persist the payload hash in the log (default is SHA2_256).

Fixes https://github.com/sigstore/rekor-tiles/issues/176

#### Release Note
NONE

#### Documentation
NONE


api/proto/dsse.proto
```diff
@@ -29,6 +29,10 @@ message DSSERequestV0_0_2 {
     io.intoto.Envelope envelope = 1 [(google.api.field_behavior) = REQUIRED];
     // All necessary verification material to verify all signatures embedded in the envelope
     repeated Verifier verifiers = 2 [(google.api.field_behavior) = REQUIRED];
+    // An optional hash algorithm to when recording the dsse payload hash in the log, by default the system will
+    // use SHA2_256
+    dev.sigstore.common.v1.HashAlgorithm payloadHashAlgorithm = 3 [(google.api.field_behavior) = OPTIONAL];
```
loosebazooka, line 34: open to other names for this

cmurphy: Maybe just `algorithm` consistent with protobuf-specs? https://github.com/sigstore/protobuf-specs/blob/95810107164a2b4c76bcdad51439f61031663078/protos/sigstore_common.proto#L116

loosebazooka: yeah I was wondering if that would be confusing? I don't think `payloadHashAlgorithm` isn't confusing. It'd be nice if it was clear that the server is going to use this, and it doesn't affect anything else they provide in the request.

pkg/types/dsse/dsse.go
```diff
@@ -65,14 +66,14 @@ func ToLogEntry(ds *pb.DSSERequestV0_0_2) (*pb.DSSELogEntryV0_0_2, error) {
...
 	return &pb.DSSELogEntryV0_0_2{
-		PayloadHash: &v1.HashOutput{
-			Algorithm: v1.HashAlgorithm_SHA2_256,
-			Digest:    payloadHash[:],
-		},
-		Signatures: signerVerifiers,
+		PayloadHash: payloadHash,
+		Signatures:  signerVerifiers,
 	}, nil
 }
```
haydentherapper, line 79: On line 101, we also need to plumb through the hash to the verifier. We also need to document that the hash specified in the request must match the hashing algorithm used with the signature algorithm.

haydentherapper: I might not have mentioned that this was the original motivation, this is to be at parity with hashedrekord, where the hash alg specified is the same as the hashing algorithm of the signature algorithm.

loosebazooka: ah I missed that.

api/proto/dsse.proto
cmurphy: :woman_shrugging:

haydentherapper: Naming isn't my strong suit, I think either would be fine. Given that it's used as part of the signature too, a more generic `algorithm` might be more accurate, but either way, they need to read the doc string.

pkg/types/dsse/dsse.go
loosebazooka: > On line 101, we also need to plumb through the hash to the verifier. We also need to document that the hash specified in the request must match the hashing algorithm used with the signature algorithm.

I think I'm confused here. Do we need the full signature spec ALSO included in here? I think there's two things at play:
1. The arbitrary log format for DSSE types. PayloadHash that helps determine if the envelope signing event was logged.
2. The "out-of-band" defined signature format ([see Sign()]( https://github.com/secure-systems-lab/dsse/blob/master/protocol.md#signature-definition))

Are we patching PublicKeyDetail from Public Key (or something embedded in the cert for certs) to the hash algorithm?

Should we just ignore the user input and parse out the hashing algorithm from the provided key?

loosebazooka:
I think this is deprecated. This information will be available by the PublicKeyDetails provided in #183

Issue #156, opened 2025-03-28
Clean up pkg/pki


cmurphy:
**Description**

- Remove hardcoded sha256 refs
- Remove support for cert chains
- @haydentherapper can add more suggestions

PR #180, opened 2025-04-03
Refactor pkg/pki into verifier library, DSSE crypto agility

haydentherapper:
I've redone the pkg/pki library to greatly simplify it, because at this point, we have no path forward for supporting any other verifiers besides certificates and public keys due to our bundle format. The changes are roughly as follows:

* Renamed PKI interface to Verifier, and changed method names
* Split x509 verifier into two, for public keys and for certificates
* Removed the Signature interface. DSSE doesn't use it because it relies on a sigstore/sigstore DSSE verification library, so it was simpler to just drop it and use the sig verification methods directly in hashedrekord
* Renamed existing verifier package to validator to clarify difference

I also fixed a bug that assumed the verifier input was PEM-encoded. The protobuf spec states that the certificate and public key RawBytes are DER-encoded.

Fixes https://github.com/sigstore/rekor-tiles/issues/156, https://github.com/sigstore/rekor-tiles/issues/176


pkg/types/verifier/verifier.go
```diff
@@ -15,27 +15,17 @@
...
-	}
-	return nil
+// Verifier represents a structure that can verify a signature, e.g. a public key or certificate
+type Verifier interface {
+	// PublicKey returns the underlying public key for signature verification
+	PublicKey() crypto.PublicKey
+	// Identities returns a list of typed keys and certificates
+	Identities() ([]identity.Identity, error)
+	// String returns a human-readable representation of the verifier, e.g. PEM-encoded
+	String() string
```
cmurphy, line 30: This is only used for the tests, can we do without?

pkg/types/verifier/verifier.go
```diff
@@ -15,27 +15,17 @@
...
-			return fmt.Errorf("missing X.509 certificate raw bytes")
-		}
-	}
-	return nil
+// Verifier represents a structure that can verify a signature, e.g. a public key or certificate
+type Verifier interface {
+	// PublicKey returns the underlying public key for signature verification
+	PublicKey() crypto.PublicKey
+	// Identities returns a list of typed keys and certificates
+	Identities() ([]identity.Identity, error)
```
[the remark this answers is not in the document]
haydentherapper: This is used by [rekor-monitor](https://github.com/sigstore/rekor-monitor/blob/9b8dadfb2ec55ebfa3265a75aebbeaada0154343/pkg/rekor/identity.go#L258) and internally for building a Rekor dataset.

pkg/types/verifier/verifier.go
haydentherapper: Similar to above, this function can be used by other libraries.

pkg/types/validator/validator.go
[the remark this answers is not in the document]
haydentherapper: Fair, I was conflicted on how best to structure this. I'll update this.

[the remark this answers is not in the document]
haydentherapper: I ended up leaving validator under pkg/types, because it was for validating the proto (and we could move the algorithm registry check under this as well). I moved the crypto verifier bits up to pkg/verifier, because as you said, they have nothing to do with the proto-defined types. The only downside is we now have pkg/verifier and pkg/verify - open to other suggestions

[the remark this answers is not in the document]
loosebazooka: This seems okay to me. Would it make sense to isolate what we want to not export out of rekor into `/internal`?

pkg/verifier/identity/identity.go
```diff
@@ -22,19 +22,11 @@ type Identity struct {
...
 	// - PKIX ASN.1 DER-encoded public key
 	// - ASN.1 DER-encoded certificate
 	Raw []byte
-	// For keys, certificates, and minisign, hex-encoded SHA-256 digest
-	// of the public key or certificate
-	// For SSH and PGP, Fingerprint is the standard for each ecosystem
-	// For SSH, unpadded base-64 encoded SHA-256 digest of the key
-	// For PGP, hex-encoded SHA-1 digest of a key, which can be either
-	// a primary key or subkey
+	// Contains hex-encoded SHA-256 digest of the public key or certificate
```
loosebazooka, line 30: a little more detail here would be helpful? SHA-256 of the raw der encoded publickey/certificate ?

pkg/verifier/verifier.go
```diff
@@ -0,0 +1,31 @@
...
+	"crypto"
+
+	"github.com/sigstore/rekor-tiles/pkg/verifier/identity"
+)
+
+// Verifier represents a structure that can verify a signature, e.g. a public key or certificate
+type Verifier interface {
+	// PublicKey returns the underlying public key for signature verification
+	PublicKey() crypto.PublicKey
+	// Identities returns a list of typed keys and certificates
```
loosebazooka, line 27: why would there be a list?

loosebazooka (commented):
looks fine to me, just some questions?

pkg/verifier/identity/identity.go
haydentherapper: added

pkg/verifier/verifier.go
haydentherapper: This was a holdover from when we supported cert chains, updated now!

pkg/types/validator/validator.go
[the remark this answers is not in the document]
haydentherapper: I'd vote to take a pass over this later on, as it's not yet clear what we need exported, to be used by sigstore-go

loosebazooka (approved):
Cool. Can address exports later

PR #183, opened 2025-04-04
Require PublicKeyDetails be specified for any verifier

haydentherapper:
The API changes are as follows:

* For `Verifier`, `PublicKeyDetails` is now required for a certificate or public key.
* For `PublicKey`, remove `PublicKeyDetails`.
* For the hashedrekord request, `HashOutput data` is now only `bytes digest`. Hash alg comes from the PublicKeyDetails.

This change will enable Rekor v2 to support signatures generated with any hash algorithm, and these key/hash algorithms can be compared against a registry of supported algorithms.


pkg/types/hashedrekord/hashedrekord.go
```diff
@@ -59,21 +56,16 @@ func ToLogEntry(hr *pb.HashedRekordRequestV0_0_2) (*pb.HashedRekordLogEntryV0_0_
...
-	case v1.HashAlgorithm_SHA2_384:
-		alg = crypto.SHA384
-	case v1.HashAlgorithm_SHA2_512:
-		alg = crypto.SHA512
-	default:
-		alg = crypto.SHA256
-	}
-	if err := sigObj.Verify(nil, keyObj, options.WithDigest(hr.Data.Digest), options.WithCryptoSignerOpts(alg)); err != nil {
+
+	// TODO: Look up hash with sigstore/sigstore's GetAlgorithmDetails(hr.Signature.Verifier.KeyDetails).GetHashType()
```
haydentherapper, line 60: Needed to undo this temporarily, we'll add this back in https://github.com/sigstore/rekor-tiles/pull/154 when it uses the AlgorithmRegistry.

haydentherapper:
@cmurphy @loosebazooka Here's the proposed changes. If you agree, we'll get this in, then update the algorithm registry PR to pull from PublicKeyDetails for determining the hash for hashedrekord. The DSSE PR can then follow by doing the same.

jku:
looks right to me, I'll let appu or colleen approve

pkg/types/hashedrekord/hashedrekord.go
cmurphy: This breaks #154 because now the algorithm is hardcoded.

Issue #123, opened 2025-03-21
Get full metrics coverage


loosebazooka:
Add in all metrics we want into `pkg/server/metrics.go` and then updated as needed everywhere

cmurphy:
Is this a duplicate of https://github.com/sigstore/rekor-tiles/issues/12 ?

Horiodino:
Well, updating the Rekor v2 server package with metrics from v1 requires selecting only those relevant to its tile-based architecture while excluding those tied to deprecated features as mentioned above.  

I read through the Rekor v2 design document and identified key metrics that align with the new architecture, such as **tile latency** and **checkpoint publication**, while marking others—like **Pub/Sub** and redundant API metrics—as obsolete.  

Based on this, the approach is to modify the provided code by **adding two essential metrics**:  
- `rekor_tile_storage_latency_summary`  
- `rekor_checkpoint_publish`  

Then, all other metrics (from the v1) should be excluded from the list to ensure alignment with Rekor v2's design (just include the present v2 and the above mentioned) Wdyt ?

haydentherapper:
We need to go through the list of metrics in https://github.com/sigstore/rekor/blob/main/pkg/api/metrics.go, and for each metric, note if we need to either a) copy it to Rekor v2, b) ignore it because it's no longer relevant, or c) do nothing because it's already in https://github.com/sigstore/rekor-tiles/blob/main/pkg/server/metrics.go.

We also should review the gRPC metrics that are provided by the go gRPC prom library and make sure they're sufficient.

I'm not sure either of those metrics would be meaningful as measured by Rekor v2. Publishing to the log happens in batches, so storage latency will be variable, and checkpoint publishing would be as well since that happens in batches. I'll file a feature request with Tessera for metrics to measure latency for the managed storage.

haydentherapper:
https://github.com/transparency-dev/trillian-tessera/issues/570 for Tessera

haydentherapper:
Here's the metrics from Rekor v1:

```
	metricNewEntries = promauto.NewCounter(prometheus.CounterOpts{
		Name: "rekor_new_entries",
		Help: "The total number of new log entries",
	})

	metricPublishEvents = promauto.NewCounterVec(prometheus.CounterOpts{
		Name: "rekor_publish_events",
		Help: "The status of publishing events to Pub/Sub",
	}, []string{"event", "content_type", "status"})

[...]

PR #184, opened 2025-04-06
Monitoring

Horiodino:
#### Summary

fixes : https://github.com/sigstore/rekor-tiles/issues/123#issuecomment-2778009093
added gRPC metrics interceptor for qps, latency, and request size  and,
a  source label to differentiate between http and direct gRPC calls

### DIFF
--- a/pkg/server/grpc.go
+++ b/pkg/server/grpc.go
@@ -34,13 +34,12 @@ import (
 	"net"
 	"os"
 	"os/signal"
-	"strings"
 	"sync"
 	"syscall"
-	"time"
 
 	pb "github.com/sigstore/rekor-tiles/pkg/generated/protobuf"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/keepalive"
@@ -59,7 +58,12 @@ func newGRPCServer(config *GRPCConfig, server rekorServer) *grpcServer {
 	var opts []grpc.ServerOption
 
 	opts = append(opts,
-		grpc.ChainUnaryInterceptor(getMetrics().serverMetrics.UnaryServerInterceptor(), customMetricsInterceptor),
+		grpc.ChainUnaryInterceptor(
+			getMetrics().serverMetrics.UnaryServerInterceptor(),
+			func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
+				return customMetricsInterceptor(ctx, req, info, handler, config)
+			},
+		),
 		grpc.ConnectionTimeout(config.timeout),
 		grpc.KeepaliveParams(keepalive.ServerParameters{MaxConnectionIdle: config.timeout}),
 		grpc.MaxRecvMsgSize(config.maxMessageSize),
@@ -134,31 +138,28 @@ func loadTLSCredentials(certFile, keyFile string) (credentials.TransportCredenti
 	return credentials.NewTLS(tlsConfig), nil
 }
 
-func customMetricsInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
-	start := time.Now()
-
+func customMetricsInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler, config *GRPCConfig) (any, error) {
 	resp, err := handler(ctx, req)
 	if err != nil {
-		slog.Error("error handling request", "error", err)
+		return resp, fmt.Errorf("grpcServiceError: %w", err)
 	}
 	m := getMetrics()
 	method := info.FullMethod
 	code := status.Code(err).String()
 
-	parts := strings.Split(method, "/")
-	service := ""
-	if len(parts) >= 2 {
-		service = parts[1]
-	}
+	service := info.FullMethod
 
-	source := "direct"
+	source := "grpc"
 	if md, ok := metadata.FromIncomingContext(ctx); ok {
-		if ua := md.Get("user-agent"); len(ua) > 0 && strings.Contains(ua[0], "grpc-gateway") {
-			source = "http"
+		if auth := md.Get(HTTPReqAuthenticatorKey); len(auth) > 0 {
+			if auth[0] == config.reqAuthenticator {
+				source = "http"
+			} else {
+				slog.Warn("invalid http-request-authenticator", "provided", auth[0])
+				return nil, status.Errorf(codes.Unauthenticated, "invalid authenticator")
+			}
 		}
 	}
-
-	latency := time.Since(start).Seconds()
 
 	reqSize := 0
 	if msg, ok := req.(proto.Message); ok {
@@ -166,7 +167,6 @@ func customMetricsInterceptor(ctx context.Context, req interface{}, info *grpc.U
 	}
 
 	m.grpcQPS.WithLabelValues(service, method, code, source).Inc()
-	m.grpcLatency.WithLabelValues(service, method, code, source).Observe(latency)
 	m.grpcRequestSize.WithLabelValues(service, method, code, source).Observe(float64(reqSize))
 
 	return resp, err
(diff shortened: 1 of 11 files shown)


pkg/server/metrics.go
```diff
@@ -41,6 +41,9 @@ type metrics struct {
 	httpLatency            *prometheus.HistogramVec
 	httpRequestsCount      *prometheus.CounterVec
 	requestSize            *prometheus.HistogramVec
```
haydentherapper, line 43: Can we rename this to httpRequestSize to differentiate from the new histogram?

pkg/server/metrics.go
```diff
@@ -82,6 +85,21 @@ var _initMetricsFunc = sync.OnceValue[*metrics](func() *metrics {
 		Help: "API Request size on HTTP calls",
 	}, []string{"code", "method"})
 
+	m.grpcLatency = f.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "rekor_latency_by_api_grpc",
```
haydentherapper, line 89:
```suggestion
		Name: "rekor_grpc_latency_by_api",
```

haydentherapper: for symmetry with the other names

pkg/server/metrics.go
```diff
@@ -82,6 +85,21 @@ var _initMetricsFunc = sync.OnceValue[*metrics](func() *metrics {
...
+		Help: "GRPC latency by service and method",
+	}, []string{"service", "method", "code", "source"})
+
+	m.grpcQPS = f.NewCounterVec(prometheus.CounterOpts{
+		Name: "rekor_qps_by_api_grpc",
+		Help: "GRPC QPS by service and method",
+	}, []string{"service", "method", "code", "source"})
+
+	m.grpcRequestSize = f.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "rekor_grpc_request_size_bytes",
```
haydentherapper, line 99:
```suggestion
		Name: "rekor_grpc_api_request_size",
```

pkg/server/metrics.go
```diff
@@ -82,6 +85,21 @@ var _initMetricsFunc = sync.OnceValue[*metrics](func() *metrics {
 		Help: "API Request size on HTTP calls",
 	}, []string{"code", "method"})
 
+	m.grpcLatency = f.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "rekor_latency_by_api_grpc",
+		Help: "GRPC latency by service and method",
+	}, []string{"service", "method", "code", "source"})
+
+	m.grpcQPS = f.NewCounterVec(prometheus.CounterOpts{
+		Name: "rekor_qps_by_api_grpc",
```
haydentherapper, line 94:
```suggestion
		Name: "rekor_grpc_qps_by_api",
```

pkg/server/grpc.go
```diff
@@ -128,3 +133,41 @@ func loadTLSCredentials(certFile, keyFile string) (credentials.TransportCredenti
 
 	return credentials.NewTLS(tlsConfig), nil
 }
+
+func customMetricsInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
```
haydentherapper, line 137:
```suggestion
func customMetricsInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
```

pkg/server/grpc.go
```diff
@@ -128,3 +133,41 @@ func loadTLSCredentials(certFile, keyFile string) (credentials.TransportCredenti
...
+	method := info.FullMethod
+	code := status.Code(err).String()
+
+	parts := strings.Split(method, "/")
+	service := ""
+	if len(parts) >= 2 {
+		service = parts[1]
+	}
+
+	source := "direct"
```
haydentherapper, line 154:
```suggestion
	source := "grpc"
```

pkg/server/grpc.go
```diff
@@ -128,3 +133,41 @@ func loadTLSCredentials(certFile, keyFile string) (credentials.TransportCredenti
...
+	if md, ok := metadata.FromIncomingContext(ctx); ok {
+		if ua := md.Get("user-agent"); len(ua) > 0 && strings.Contains(ua[0], "grpc-gateway") {
```
haydentherapper, line 156: I couldn't find any info online about gRPC Gateway setting the user agent, but even if so, this value could come from a user too and manipulate metrics.

Can we add a secret value that only the server knows and add this to requests to gRPC from the gateway? Something like:

```
// add new flag in cmd/rekor-server/app/serve.go
serveCmd.Flags().String("http-request-authenticator", "", "shared secret value to authenticate requests between HTTP and gRPC")

// modify the server mux in pkg/server/http.go
func setHTTPRequestAuthenticator(_ context.Context, _ *http.Request) metadata.MD {
	return metadata.Pairs(HTTPReqAuthenticatorKey, config.ReqAuthenticator)
}

runtime.NewServeMux(runtime.WithMetadata(setSharedRequestSeceret), ...)
```

If it's set and correct, then we know it comes from the gateway. If it's set and is wrong, then the request should be rejected. Otherwise, it is an external request.

I'm open to ideas of a simpler solution as well. Another option is randomly generating a value on server startup, but this wouldn't work if a private deployer ran their own proxy instead of gRPC gateway and still wanted this feature.

haydentherapper (changes requested):
Thanks for working on this!

pkg/server/metrics.go
```diff
@@ -41,6 +41,9 @@ type metrics struct {
 	httpLatency            *prometheus.HistogramVec
 	httpRequestsCount      *prometheus.CounterVec
 	requestSize            *prometheus.HistogramVec
+	grpcLatency            *prometheus.HistogramVec
+	grpcQPS                *prometheus.CounterVec
+	grpcRequestSize        *prometheus.HistogramVec
 }
```
haydentherapper, line 47: Should we also add httpQPS?

haydentherapper (changes requested):

pkg/server/grpc.go
```diff
@@ -128,3 +133,41 @@ func loadTLSCredentials(certFile, keyFile string) (credentials.TransportCredenti
...
+	}
+	m := getMetrics()
+	method := info.FullMethod
+	code := status.Code(err).String()
+
+	parts := strings.Split(method, "/")
+	service := ""
+	if len(parts) >= 2 {
+		service = parts[1]
+	}
```
loosebazooka, line 152: I'm not fond of this service parsing code? Especially since it implies we *could* end up with a `""` service string.
Why not just keep the `FullMethod` string? What does trimming the package do for us?

pkg/server/grpc.go
```diff
@@ -128,3 +133,41 @@ func loadTLSCredentials(certFile, keyFile string) (credentials.TransportCredenti
 
 	return credentials.NewTLS(tlsConfig), nil
 }
+
+func customMetricsInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	start := time.Now()
+
+	resp, err := handler(ctx, req)
```
loosebazooka, line 140: I think we need to better indicate that this err should be returned. A name like grpcServiceErr makes more sense to me.

pkg/server/metrics.go
```diff
@@ -82,6 +85,21 @@ var _initMetricsFunc = sync.OnceValue[*metrics](func() *metrics {
 		Help: "API Request size on HTTP calls",
 	}, []string{"code", "method"})
 
+	m.grpcLatency = f.NewHistogramVec(prometheus.HistogramOpts{
```
loosebazooka, line 88: Are we not already getting a lot of these from the default server metrics?

pkg/server/grpc.go
haydentherapper: The UUID would be simpler. Only downside is what I mentioned, that it assumes the gRPC and HTTP servers are always run together. That's probably reasonable?

pkg/server/metrics.go
loosebazooka: I think we can enable this? https://github.com/grpc-ecosystem/go-grpc-middleware/blob/0e172e476feda43f0e83cc894fe2d636c1d820fa/providers/prometheus/server_options.go#L37

Horiodino:
Is this PR still valid? I noticed that @loosebazooka is now assigned to the related issue. Let me know if we should go ahead and close this one, and I can shift  to other issues.

Issue #38, opened 2025-02-24
Create a framework to run a local Rekor v2 instance that can be used for development and CI testing in clients


haydentherapper:
**Description**

Evaluate using either:

* [Cosign's e2e test suite](https://github.com/sigstore/cosign/tree/main/test)
* [Scaffolding CI action](https://github.com/sigstore/scaffolding/tree/main/actions/setup)
* [sigstore-js mocks](https://github.com/sigstore/sigstore-js/tree/main/packages/mock)

Hayden's 2c: Cosign is the simplest solution and completely comprehensive. Scaffolding requires kind and relies on K8s configuration that is not maintained to the same level as sigstore/helm-charts.

haydentherapper:
From standup, [how Cosign stands up the TSA](https://github.com/sigstore/cosign/blob/37bae90768f66c930b5630d0f570778141878737/test/e2e_test.go#L1353-L1358), but this can be replaced with the container

PR #191, opened 2025-04-07
compose.yml: Tweak the ports

jku:
This is only a quality of life tweak: select ports that do not clash with ports used by sigstore/rekor so that both projects can be composed.

-- 

I think this will also make #38 slightly easier to get going: we likely will end up maintaining the required compose files in our sigstore-in-a-box itself but using the existing compose files lets us get started easily.


loosebazooka (approved):
I don't know if we care about exposing the metrics port, I think I added that so I could actually manually see what was going on. But I guess it doesn't really matter.

Issue #173, opened 2025-04-02
Have cloudsdk image references for dependabot add a version comment


loosebazooka:
See: https://github.com/sigstore/rekor-tiles/pull/164#discussion_r2020563201

PR #199, opened 2025-04-08
Update to a multi-platform versioned gcloud container

haydentherapper:
Also fixes e2e testing and `make all` on macOS, and fixes linting errors for Dockerfiles.

It appears that too many concurrent requests on macOS causes connections to get dropped. Not sure why, but a simple retry loop seems to resolve it.

Fixes https://github.com/sigstore/rekor-tiles/issues/173

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.24.2@sha256:991aa6a6e4431f2f01e869a812934bd60fbc87fb939e4a1ea54b8494ab9d2fc6 AS builder
+FROM golang:1.24.2@sha256:fb224f950b0dfb889f8f1122bf3dfc21976735ccf983b73a1e6e014215a272f5 AS builder
 ENV APP_ROOT=/opt/app-root
 ENV GOPATH=$APP_ROOT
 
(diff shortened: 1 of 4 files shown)


jku (dismissed):
Adding `time.Sleep()` to make tests pass is rarely a good sign but maybe this is fine? At least it shouldn't usually make things slower since we'll anyway wait for the checkpoint.

haydentherapper:
I can remove looping with wait groups to limit concurrent requests, lemme do that.

haydentherapper:
@loosebazooka @jku found a much simpler solution, you can set a limit on the number of concurrent requests for an errgroup

Dockerfile
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.24.2@sha256:991aa6a6e4431f2f01e869a812934bd60fbc87fb939e4a1ea54b8494ab9d2fc6 AS builder
+FROM golang:1.24.2@sha256:fb224f950b0dfb889f8f1122bf3dfc21976735ccf983b73a1e6e014215a272f5 AS builder
```
cmurphy, line 16: Can't tell what this hash change means, is this a newer or older version? It's not the latest dockerhub version of 1.24.2.

haydentherapper: It's been updated again, this was the hash yesterday. :| Will update, `crane digest golang:1.24.2` now returns `sha256:1ecc479bc712a6bdb56df3e346e33edcc141f469f82840bab9f4bc2bc41bf91d`

loosebazooka (approved):

Issue #198, opened 2025-04-08
Create workflows for releasing production binary and container


haydentherapper:
**Description**

This tracks creating a workflow for releasing prebuilt binaries for a few architectures. There's a lot of examples we'll pull from from other Sigstore repos.

For the container build, we'll likely use Ko. The container base image will be Distroless, won't include dlv, and the binary will include the various ldflags for versioning.

PR #200, opened 2025-04-08
Add release scripts for production binaries and containers 

haydentherapper:
This adds a goreleaser script for publishing binaries. The binaries will
be indirectly signed from the GHA provenance.

Also updates the existing container release script to treat tags and
pushes separately, using a different Dockerfile with a minimal base
image.

Fixes https://github.com/sigstore/rekor-tiles/issues/198

Signed-off-by: Hayden B <8418760+haydentherapper@users.noreply.github.com>

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -13,7 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.24.2@sha256:1ecc479bc712a6bdb56df3e346e33edcc141f469f82840bab9f4bc2bc41bf91d AS builder
+# Debian 12 (Bookworm) image to build binary for distroless/static-debian12
+FROM golang:1.24.2-bookworm@sha256:00eccd446e023d3cd9566c25a6e6a02b90db3e1e0bbe26a48fc29cd96e800901 AS builder
 ENV APP_ROOT=/opt/app-root
 ENV GOPATH=$APP_ROOT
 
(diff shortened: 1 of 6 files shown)


haydentherapper:
Depends on https://github.com/sigstore/rekor-tiles/pull/199, will rebase.

haydentherapper:
@cpanato Can you take a look? I took inspiration from timestamp-authority, but tried to avoid outright copying every line to make sure this was minimal and up to date. Main diff is dropping cosign in favor of just using the GHA provenance, not using Ko (could simplify this later, but I liked being explicit about each step), and relying on Make to provide the ldflags rather than have a separate script.

Dockerfile.release
```diff
@@ -0,0 +1,40 @@
...
+# Add source code
+ADD ./cmd/ $APP_ROOT/src/cmd/
+ADD ./pkg/ $APP_ROOT/src/pkg/
+
+ARG SERVER_LDFLAGS
+RUN go build -ldflags "${SERVER_LDFLAGS}" ./cmd/rekor-server
+RUN CGO_ENABLED=0 go build -gcflags "all=-N -l" -ldflags "${SERVER_LDFLAGS}" -o rekor-server ./cmd/rekor-server
+
+# Multi-Stage production build
+FROM gcr.io/distroless/static-debian12:nonroot@sha256:c0f429e16b13e583da7e5a6ec20dd656d325d88e6819cafe0adb0828976529dc AS deploy
```
cmurphy, line 34: Could we just update the original dockerfile to have this minimal base? These dockerfiles are almost the same, it seems like we could just use the same one. Otherwise we'll have to be conscious of keeping them in sync with each other. It seems like we don't bother with separate dockerfiles for the other services?

haydentherapper: For other services, we've had one ko file which is for production containers and one Dockerfile only for development. I'm fine with ko but have a preference towards being explicit with the steps for our container builds, which is why there are now two dockerfiles.

There is a difference between these, as the other dockerfile has a third stage of installing dlv. We could add an argument to skip this stage when set. WDYT?

cmurphy: Having an argument to skip the debug stage sgtm

haydentherapper: Updated, lemme know what you think! We do need different base images for the prod vs debug containers since the debug container needs a shell, so I added multiple stages.

Dockerfile
```diff
@@ -26,22 +30,26 @@ ADD ./cmd/ $APP_ROOT/src/cmd/
...
 
+# Multi-stage deployment build
+FROM gcr.io/distroless/static-debian12:nonroot@sha256:c0f429e16b13e583da7e5a6ec20dd656d325d88e6819cafe0adb0828976529dc AS deploy
 # Retrieve the binary from the previous stage
 COPY --from=builder /opt/app-root/src/rekor-server /usr/local/bin/rekor-server
 
+# Multi-stage debugger build
+FROM golang:1.24.2@sha256:18a1f2d1e1d3c49f27c904e9182375169615c65852ace724987929b910195b2c AS debug
+# Copy debugger server
+COPY --from=builder /opt/app-root/src/rekor-server_debug /usr/local/bin/rekor-server
```
cmurphy, line 47: I tried building this with

```
docker build . --build-arg IMAGE_VARIANT=deploy
```
and it failed at this step. It's missing the debug binary built on line 37. I'm not sure if the Dockerfile syntax is smart enough to only run certain steps depending on the build arguments.

If it's not easy to add conditionals in one dockerfile I'm fine with splitting it in two.

haydentherapper: Debugging this now, it looks to be due to building with CGO enabled, which will fail on distroless. I've got a fix, pushing it up now

haydentherapper: Eh, now the build times out. Lemme just switch base images.

haydentherapper: Oh I just reread your statement, yep I see the problem.

haydentherapper:
The issue is that the distroless image doesn't include `curl` so the healthcheck fails (slightly). Talking with Appu now for ideas...

haydentherapper:
To keep the container small, not going to copy over wget or curl. We have this same base image for rekor/fulcio and haven't had any issues deploying with healthchecks, so I don't think this is necessary.

cmurphy (approved):

PR #212, opened 2025-04-10
Add panic interceptor and metrics for it

loosebazooka:
#### Summary
Handles when grpc processing panics

ref: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/examples/server/main.go#L111

### DIFF
--- a/pkg/server/grpc.go
+++ b/pkg/server/grpc.go
@@ -33,8 +33,13 @@ import (
 	"net"
 	"os"
 	"os/signal"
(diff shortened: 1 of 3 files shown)


haydentherapper (approved):
Do we need one for HTTP as well? The bulk of the server logic is in gRPC but there could still be a panic before or after the gRPC call,

loosebazooka:
Looks like http.Servers default serving loops handles panic as 500s automatically.

PR #214, opened 2025-04-10
Specify correct hash algorithm in response

### LINKED ISSUE #195: Response should have the correct hash algorithm
**Description**

With the crypto agility work, the request specifies the key details, which include the sig alg and digest alg. In the response, we specify the proto-defined hash alg for convenience. We need a way to convert key details to the proto hash alg, as we currently only have a way to get crypto.Hash

haydentherapper:
This fixes a bug for hashedrekord where the response hash algorithm was always hardcoded to SHA-256. For DSSE, the fix was a bit more tricky. Because we may have multiple PublicKeyDetails with different hash algorithms, because a DSSE can have multiple signatures, we need to select one of the algorithms to use to hash the payload.

We could have hardcoded SHA-256, but this is a step backwards for crypto agility. We could reject requests with different hash algorithms, but this would prevent artifacts signed with classical and PQ crypto algorithms. We could outright reject DSSEs with multiple sigs, but this is likely to have a larger impact. Therefore, we implemented a simple selection algorithm to pick the strongest and newest digest alg.

Fixes #195


pkg/algorithmregistry/algorithmregistry.go
```diff
@@ -33,6 +35,10 @@ var (
 		v1.PublicKeyDetails_PKIX_ED25519,
 		v1.PublicKeyDetails_PKIX_ED25519_PH,
 	}
+	// Opinionated ordering of hash algorithms from oldest and weakest to newest and strongest.
+	// Must be updated if AllowedClientSigningAlgorithms is updated with a new digest.
```
loosebazooka, line 39: I think we should add a comment above at AllowedClientSigningAlgorithms to update this.

pkg/types/dsse/dsse.go
```diff
@@ -102,46 +116,50 @@ func verifyEnvelope(verifiers map[*pb.Verifier]verifier.Verifier, pbenv *pbdsse.
...
 			break // if all signatures have been verified, do not attempt anymore
 		}
 
 		algDetails, err := signature.GetAlgorithmDetails(v.KeyDetails)
 		if err != nil {
-			return nil, fmt.Errorf("getting key algorithm details: %w", err)
+			return nil, nil, fmt.Errorf("getting key algorithm details: %w", err)
 		}
 		alg := algDetails.GetHashType()
+		dsseHashAlgSet[alg] = algDetails.GetProtoHashType()
```
loosebazooka, line 131: oh yeah, I dunno if this is ideal, or we should've created the algorithm mapping separately. This feels a bit goofy, but I can live with it.

loosebazooka (approved):

pkg/algorithmregistry/algorithmregistry.go
haydentherapper: Added a comment.

pkg/types/dsse/dsse.go
haydentherapper: Yea, given it's for this one special case, I'm fine with this. Multi-sigs types aren't the norm (if we were to add more).

haydentherapper:
Hm, signature ordering might not be canonicalized since this test is failing sporadically. Looking into it...

loosebazooka (dismissed):

haydentherapper:
There is a bug, there's no canonicalized ordering for the sigs, adding that in now.

loosebazooka:
Oh I guess you gotta fix that test.

pkg/algorithmregistry/algorithmregistry.go
```diff
@@ -78,3 +88,22 @@ func CheckEntryAlgorithms(pubKey crypto.PublicKey, alg crypto.Hash, algorithmReg
...
 }
+
+// SelectHashAlgorithm picks the newest and strongest hash algorithm.
+func SelectHashAlgorithm(algs []crypto.Hash) (crypto.Hash, error) {
+	if len(algs) == 0 {
+		return crypto.Hash(0), fmt.Errorf("hash algorithm slice is empty")
+	}
+	// sort from weakest to strongest
+	// unknown algorithms will be sorted first in the array
+	sort.Slice(algs, func(i, j int) bool {
```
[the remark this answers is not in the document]
haydentherapper: The problem is that slices.Index returns -1 if not found. If the opinionated list is ordered strongest to weakest, I can't use `<` as a comparison since unknown algs will be in the front. If I use `>`, then the opinionated list must be ordered weakest to strongest, which I think is inconsistent to have different sorting orderings.

haydentherapper:
Sig ordering is now canonicalized.

pkg/types/dsse/dsse.go
```diff
@@ -75,86 +76,112 @@ func ToLogEntry(ds *pb.DSSERequestV0_0_2, algorithmRegistry *signature.Algorithm
...
 		}
 	}
-	signerVerifiers, err := verifyEnvelope(verifiers, ds.Envelope, algorithmRegistry)
+	signerVerifiers, dsseHashAlgs, err := verifyEnvelope(verifiers, ds.Envelope, algorithmRegistry)
 	if err != nil {
 		return nil, err
 	}
-	payloadHash := sha256.Sum256(ds.Envelope.Payload)
+
+	// Canonicalize the order of the signatures
```
loosebazooka, line 84: I think it would be helpful to define canonicalization algorithm here.

loosebazooka (approved):

pkg/types/dsse/dsse.go
haydentherapper: Updated in https://github.com/sigstore/rekor-tiles/pull/220/

PR #216, opened 2025-04-11
Measure request payload sizes

loosebazooka:
Uses grpc stats to get measurement information on payload sizes

#### Summary
Interceptors don't provide wiresize of payloads. StatsHandler appears to be the right place to get this info, so we grab it from there.

### DIFF
--- a/pkg/server/grpcstats.go
+++ b/pkg/server/grpcstats.go
@@ -1,0 +1,70 @@
+// Copyright 2025 The Sigstore Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"context"
+	"strings"
+
+	"google.golang.org/grpc/stats"
+)
+
+// GrpcStatsHandler consumes grpc stats and converts them to metrics
+type GrpcStatsHandler struct{}
+
+func (st *GrpcStatsHandler) TagConn(ctx context.Context, _ *stats.ConnTagInfo) context.Context {
+	// this is a no-op we don't care about connections for now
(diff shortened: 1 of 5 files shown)


pkg/server/metrics.go
```diff
@@ -68,6 +71,13 @@ var _initMetricsFunc = sync.OnceValue[*metrics](func() *metrics {
 		Help: "The total number of new dsse log entries",
 	})
 
+	// grpc_packet_part should always be "payload" but we can measure "header" or "trailer" in the future
+	// if we so desire
+	m.grpcRequestSize = f.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "grpc_api_request_size",
```
haydentherapper, line 77: name should start with rekor for consistency

pkg/server/grpcstats.go
```diff
@@ -0,0 +1,70 @@
...
+
+type rpcStatCtxKey struct{}
+
+func (st *GrpcStatsHandler) TagRPC(ctx context.Context, info *stats.RPCTagInfo) context.Context {
+	// add the rpc info to the context so we can access the method name later
+	return context.WithValue(ctx, rpcStatCtxKey{}, info)
+}
+
+// HandleRPC processes the RPC stats. Note: All stat fields are read-only.
+func (st *GrpcStatsHandler) HandleRPC(ctx context.Context, stat stats.RPCStats) {
```
haydentherapper, line 44: Could we add a test in the e2e tests that checks this is working as intended?

haydentherapper (commented):
Good finds. Seems like a good feature request for the middleware to support specifying additional metrics to initialize

pkg/server/grpcstats.go
loosebazooka: That it's populating measurements?

pkg/server/metrics.go
loosebazooka: this is actually consistent with the stats produced by grpc's default server metrics set.
see: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/providers/prometheus/server_metrics.go#L33

pkg/server/grpcstats.go
haydentherapper: Yea, just looking for some way to get code coverage for what's added here. Not going to block the PR on this though. Some e2e testing for all metrics would be nice.

haydentherapper (approved):

loosebazooka:
https://github.com/sigstore/rekor-tiles/issues/224

PR #243, opened 2025-04-20
Prevent concurrent releases

haydentherapper:
This prevents two concurrent releases or container builds from overwriting one another.

Signed-off-by: Hayden B <haydentherapper@users.noreply.github.com>

### DIFF
--- a/.github/workflows/build_container.yml
+++ b/.github/workflows/build_container.yml
@@ -30,6 +30,11 @@ env:
 
 permissions: {}
 
+# Prevent concurrent runs of this workflow, but do not cancel any in-progress
+concurrency:
+  group: "build-container"
+  cancel-in-progress: false
+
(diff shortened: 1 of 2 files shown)


.github/workflows/build_container.yml
```diff
@@ -30,6 +30,11 @@ env:
 
 permissions: {}
 
+# Prevent concurrent runs of this workflow, but do not cancel any in-progress
+concurrency:
```
loosebazooka, line 34: how does this end up working? Just queuing up releases?

loosebazooka (approved):

.github/workflows/build_container.yml
haydentherapper: Yep exactly, it'll queue them up in order of PR merges. The reason I did this is so that `latest` is always the last merged commit, otherwise there's a race for which build ends last.

PR #252, opened 2025-04-21
Add protos as target for make all

haydentherapper:
Without this, make clean && make all fails since the protos have been deleted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 
 .PHONY: all test clean lint gosec ko-local tools ldflags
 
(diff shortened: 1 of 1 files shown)


loosebazooka (dismissed):
I worry that this makes the build expensive? I don't know that I exactly did a good job of writing the makefile def for `protos`

haydentherapper:
We aren't calling "make all" anywhere currently, this is just for development. Also we'll be changing this anyways once the protos are moved.

Issue #40, opened 2025-02-24
Develop a sharding playbook


haydentherapper:
**Description**

i.e. instantiating a new tree and rolling out the new log prefix and public key to clients. Requires clients to support SigningConfig.

haydentherapper:
Blocked by https://github.com/sigstore/rekor-tiles/issues/63

haydentherapper:
Following https://github.com/transparency-dev/trillian-tessera/issues/589 - This may lower costs by sharing a Spanner instance

Issue #46, opened 2025-02-24
Create prober tests for the new Rekor


haydentherapper:
**Description**

PR #254, opened 2025-04-21
Create initial sharding playbook

haydentherapper:
This includes the sharding playbook to be used by oncallers to spin up a new shard.

There's a TODO to add a dashboard link once we have monitoring, and WIPs as we are actively developing the prober and figuring out shard turn down.

Towards https://github.com/sigstore/rekor-tiles/issues/40

### DIFF
--- a/SHARDING.md
+++ b/SHARDING.md
@@ -1,0 +1,454 @@
+# Sharding Playbook
+
+This playbook walks through the steps to create a new log shard
+for Rekor v2.
+
+Roughly, the steps are:
+
+1. Create the GCP resources with Terraform
+2. Generate key material
+3. Spin up the new infrastructure with Helm and Terraform
+4. Verify the health of the infrastructure with manual testing
+5. Run prober tests
+6. Update TrustedRoot with the new key material and SigningConfig with the new shard URL
+7. Turn down the old shard
+
+## Terraform Module Changes
+
+New shards are declared as additional modules under the respective
+environment. This is a different pattern than other modules, which are all
+initialized through the
+[`sigstore` module](https://github.com/sigstore/terraform-modules/tree/main/gcp/modules/sigstore).
+This design allowed deployers to initialize this one module and set all required variables, at
+which point you'd have a complete Sigstore stack on GCP. The downside of this approach is
+private deployers that don't need all GCP infrastructure have to initialize all resources,
+though deployers can always initialize each module referenced in `sigstore` individually.
+
+The other downside to referencing the `sigstore` module is that an update to a sub-module
+will concurrently update all deployed shards.  Declaring each shard as its own module allows us
+to make updates to shards one-by-one.
+
+The tiled transparency log module we'll reference in Terraform is
+[`tiles_tlog`](https://github.com/sigstore/terraform-modules/tree/main/gcp/modules/tiles_tlog).
+
+## Steps
+
+### Temporarily Remove Org Restriction
+
+We are restricted from creating public buckets, which are required for
+the storage backend for Rekor v2 and Tessera. Temporarily disable
+the `iam.allowedPolicyMemberDomains` constraint.
+
+[Example PR](https://github.com/sigstore/public-good-instance/pull/2938)
+
+For production, update the `projects_project-rekor` module. For staging,
+update the `projects_projectsigstore-staging` module. These modules come from
+[this file](https://github.com/sigstore/public-good-instance/blob/main/iam/resource_hierarchy/main.tf).
+
+### Create GCP resources
+
+To add a new shard, create a module under either
+[`staging.tf`](https://github.com/sigstore/public-good-instance/blob/main/terraform/environments/staging/1-infrastructure/staging.tf)
+or
+[`production.tf`](https://github.com/sigstore/public-good-instance/blob/main/terraform/environments/production/1-infrastructure/production.tf).
+
+[Example PR](https://github.com/sigstore/public-good-instance/pull/2907/). For the example,
+you'll only need to look at `staging.tf` and `general.auto.tfvars`. Remember to update
+the correct environment for production or staging.
+
+Modules should be named based on the year and how many shards
+have been created in the year. For example, the module should
+be named `tiles_tlog_log2026_1` for the first shard in 2026. We'll
+use `log2026-1` as the name of the shard itself for this document.
+
+Update `general.auto.tfvars` to include the shard variable values:
+
+```
+rekor_tiles_shards = {
+  "log2026-1" = {
+    cluster_namespace_suffix                = "rekor-tiles-system"
+    bucket_name_suffix                      = "sigstore-dev-rekor-tiles"
+    spanner_processing_units                = 100
+    spanner_instance_name_suffix            = "rekor-tiles"
+    spanner_instance_display_name_suffix    = "rekor-tiles"
+    network_endpoint_group_http_name_suffix = "rekor-tiles-neg"
+    network_endpoint_group_grpc_name_suffix = "rekor-tiles-neg-grpc"
+    // network_endpoint_group_zones = ["us-central1-c", "us-central1-f", "us-central1-b"] 
+  }
+}
+```
+
+Note that `network_endpoint_group_zones` needs to remain commented out.
+Terraform will error out trying to discover the network endpoint group (NEG) resources
+if we declare the zones but the NEGs haven't been created by GKE through the Helm chart yet.
+
+Update `production.tf` or `staging.tf` based on the example. Make sure to update
+the `module` name, `shard_name` variable, and the name of the map key for `var.rekor_tiles_shards`.
+
+After merging, run `terraform plan` through the GitHub Actions workflow
+for [staging](https://github.com/sigstore/public-good-instance/actions/workflows/env-staging.yml)
+or [prod](https://github.com/sigstore/public-good-instance/actions/workflows/env-prod.yml).
+Confirm the expected resources are created, then run `terraform apply` through the workflow.
+Note this will also remove the org restriction.
+
+### Create Key Material
+
+The public-good instance uses a Tink keyset to sign checkpoints. Tink
+encrypts a signing key using a "key encryption key" (KEK) managed by GCP KMS,
+and loads the key into memory on service startup.
+
+Tink is beneficial for reliability and cost reduction - KMS is now only
+a startup dependency rather than a runtime dependency, and signing happens
+exclusively in memory.
+
+To create a shard KEK:
+
+1. Grant yourself access to use the new KEK. Update the `staging.tf`
+or `production.tf` config to include the following IAM resource. The shard
+keyring name will be `<shard name>-rekor-tiles-keyring`, e.g. `log2026-1-rekor-tiles-keyring`.
+
+```
+resource "google_kms_key_ring_iam_member" "rekor-tiles-keyring" {
+  key_ring_id = "<project-rekor or projectsigstore-staging>/global/<name of shard keyring>"
+  role        = "roles/cloudkms.cryptoOperator"
+
+  member = "user:<username>@sigstore.dev"
+}
+```
+
+Request a review from an infrastructure maintainer and merge.
+
+[Example PR](https://github.com/sigstore/public-good-instance/pull/2942)
+
+2. Run `terraform plan` and `terraform apply` using the GHA workflow.
+
+3. [Install the gcloud CLI](https://cloud.google.com/sdk/docs/install) if not already installed.
+
+4. Authenticate to gcloud with application credentials: `gcloud auth application-default login`
+
+5. Using [tinkey](https://developers.google.com/tink/tinkey-overview#installation), create an encrypted Tink keyset:
+
+```
+tinkey create-keyset --key-template ED25519 --out enc-keyset.cfg --master-key-uri gcp-kms://projects/<project name>/locations/<region>/keyRings/<key-ring>/cryptoKeys/checkpoint-signer-key-encryption-key
+```
+
+Example:
+
+```
+tinkey create-keyset --key-template ED25519 --out enc-keyset.cfg --master-key-uri gcp-kms://projects/projectsigstore-staging/locations/global/keyRings/log2026-1-rekor-tiles-keyring/cryptoKeys/checkpoint-signer-key-encryption-key
+```
+
+TODO: Once https://github.com/sigstore/scaffolding/pull/1627 is merged, use `create-tink-keyset` rather than
+`tinkey` since this command doesn't require a Java runtime environment and will output the public key.
+
+Save the output file `enc-keyset.cfg`, which will be used as a value in the Helm chart.
+
+You'll need the public key from the private keyset for the TUF TrustedRoot later on.
+Unfortunately, Tinkey does not provide a utility to transform either a public or private keyset
+into a PEM-encoded public key. You'll fetch this public key from the server logs.
+
+6. Revoke IAM access by reverting the PR and removing `google_kms_key_ring_iam_member`, and run `plan` and `apply`.
+
+### Create Kubernetes namespace
+
+Each shard is isolated in its own Kubernetes namespace. For either
+[production](https://github.com/sigstore/public-good-instance/blob/main/terraform/environments/production/helm-charts-values/argocd-apps.yaml)
+or
+[staging](https://github.com/sigstore/public-good-instance/blob/main/terraform/environments/staging/helm-charts-values/argocd-apps.yaml),
+add a destination namespace:
+
+```
+projects:
+  utilities:
+    ...
+    destinations:
+      ...
+      - namespace: log2026-1-rekor-tiles-system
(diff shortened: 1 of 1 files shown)


SHARDING.md
```diff
@@ -0,0 +1,254 @@
+# Sharding Playbook
+
+This playbook walks through the steps to create a new log shard.
+The steps should be the same for Rekor and the CT log, though the
+CT process for generating key material may differ.
```
loosebazooka, line 5: rekor shouldn't know about CT, or this should be in another repo?

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+
+Roughly, the steps are:
+
+1. Create the GCP resources with Terraform
```
loosebazooka, line 9: same here: this seems specific to our public good instance.

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+environment. This is a different pattern than other modules, which are all
+initialized through the `sigstore` module. Declaring each shard as its
+own module allows us to make updates to shards one-by-one. Otherwise,
+by updating the module reference of `sigstore`, all shards would be
+concurrently updated.
+
+To add a new shard, create a module under either
+`terraform/environments/staging/1-infrastructure/staging.tf` or
+`terraform/environments/production/1-infrastructure/production.tf`.
+Name the module `rekor-tiles-log-<year>-<revision>`, e.g.
```
loosebazooka, line 29: what does revision mean here? That we can have multiple shards per year? Should they always count up?

nvm, somewhat answered in the example below, but some guidance on revision would still be helpful.

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+6. Update prober
+4. Perform manual testing
+5. Update TrustedRoot with the new key material and SigningConfig with the new shard URL
+6. Turning down the old shard
+
+## Create GCP resources
+
+New shards are declared as additional modules under the respective
+environment. This is a different pattern than other modules, which are all
+initialized through the `sigstore` module. Declaring each shard as its
```
loosebazooka, line 21: I'm not sure exactly what this means (probably because I have no context into infra). Why are other modules initialized through the single `sigstore` module?

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+own module allows us to make updates to shards one-by-one. Otherwise,
+by updating the module reference of `sigstore`, all shards would be
+concurrently updated.
+
+To add a new shard, create a module under either
+`terraform/environments/staging/1-infrastructure/staging.tf` or
+`terraform/environments/production/1-infrastructure/production.tf`.
```
loosebazooka, line 28: can these be links to somewhere?

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+encrypts a signing key using a "key encryption key" managed by GCP KMS,
+and loads the key into memory on service startup.
+
+Tink is beneficial for reliability and cost reduction - KMS is now only
+a startup dependency rather than a runtime dependency, and signing happens
+exclusively in memory.
+
+To create a shard signing key:
+
+1. Grant yourself access to use the new signing key. Update the `staging.tf`
```
loosebazooka, line 67: you can just do this without someone approving it?

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+
+Update `argocd/bootstrap-utilities/staging/values.yaml` or `argocd/bootstrap-utilities/production/values.yaml`
+to include the new shard. Under `rekor-tiles.shards`, copy the previous shard instantiation, and update
+the values accordingly based on the Terraform resources names. For `keyset`, copy the contents of
+`enc-keyset.cfg` (Since this value is encrypted, it's secure to include it in the config rather than
+use GCP Secrets).
+
+<!--- TODO: Add code pointers for values -->
+
+Merge, and wait for ArgoCD to create the resources and spin up the service.
```
loosebazooka, line 101: wait how long? is there some indicator? maybe an an oncall person would already know this?

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Update Prober
+
+Update the prober configuration for
```
loosebazooka, line 105: does prober work with v2?

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+
+<!--- TODO: Update prober to support testing against multiple Rekor shard URLs -->
+<!--- TODO: Update prober to test reads and writes separately, since frozen shards will still support reads -->
+
+## Perform Manual Testing
+
+Using Cosign, test signing and verification against this new shard. You will need to manually update the SigningConfig
+and TrustedRoot files and pass them via CLI flags since those files are pulled from TUF and will not yet have been updated.
+
+<!--- TODO: Update with example TrustedRoot and SigningConfig -->
```
loosebazooka, line 119: include example call to cosign using custom trustedroot and signingconfig.

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+can verify proofs from the shard and use the shard during signing.
+
+For TrustedRoot, the validity window for the shard should be from the point
+when the shard was spun up. For SigningConfig, the validity window should start
+after the TUF timestamp expires to guarantee all clients will have picked up
+the latest TrustedRoot. In practice for our TUF root with timestamp validity
+of one week, this means setting the validity window to be at least one week
+past when the TUF targets are signed.
+
+You will need to estimate two timestamps:
```
loosebazooka, line 134: I think we should have a well defined overlap period. maybe 1 day/week/month or whatever?

Having to guess this means people will ask around to get a recommendation instead of us just having this decision already made.

SHARDING.md
haydentherapper: Will update, just wanted to state that the Rekor and CT sharding processes should be the same.

SHARDING.md
haydentherapper: Yep, I'll move this playbook to PGI.

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+Merge and wait for ArgoCD to shut down the server.
+
+### Remove GCP Resources
+
+We only need a subset of GCP resources to serve read traffic. To save costs,
+we can turn down the database and destroy the signing key.
+
+<!--- TODO: Specify how we should do this -->
+
+<!--- TODO: Specify how to convey to monitors the shard is frozen.
```
haydentherapper, line 252: One other thought - For CT, we can't add extension lines, so we if we want to denote a frozen shard in the same way, baking in the checkpoint might be the only viable approach.

haydentherapper (commented):
Thanks for the feedback, will work on updating!

SHARDING.md
haydentherapper: Tracking in https://github.com/sigstore/rekor-tiles/issues/46

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+when the shard was spun up. For SigningConfig, the validity window should start
+after the TUF timestamp expires to guarantee all clients will have picked up
+the latest TrustedRoot. In practice for our TUF root with timestamp validity
+of one week, this means setting the validity window to be at least one week
+past when the TUF targets are signed.
+
+You will need to estimate two timestamps:
+
+* New shard's start for SigningConfig, which should be one week past when the TUF
+timestamp will be signed. Overestimate if needed, as you don't want clients to start
```
jku, line 137: Would this be more understandable to folks with no knowledge of root-signing details?

```suggestion
* New shard's start for SigningConfig, which should be one week past when the root-signing PR is merged and new TUF repository is published. Overestimate if needed, as you don't want clients to start
```

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+using the new shard before all clients have picked up the latest TrustedRoot.
+* Previous shard's end for TrustedRoot, which is when the previous shard's write path
+is turned off. This must be after the start of the new shard. It doesn't have to
```
jku, line 140:
```suggestion
* Previous shard's end for TrustedRoot: clients will consider this shards entry timestamps invalid if they are after this date. This must be after the start of the new shard. It doesn't have to
```

I don't know if the suggestion is any better -- you could reword the original as well to make it a bit clearer (that this date doesn't need to be the moment shard is turned off, just needs to be later than the turn off date.

SHARDING.md
```diff
@@ -0,0 +1,254 @@
...
+    // previous shard
+    {
+      "baseUrl": "...",
+      "hashAlgorithm": "SHA2_256",
+      "publicKey": {
+        "rawBytes": "...",
+        "keyDetails": "...",
+        "validFor": {
+          "start": "...",
+          "end": "<UTC timestamp for approx when the shard will be shut down>"
```
jku, line 173:
```suggestion
          "end": "<UTC timestamp that is after the shard is turned read only>"
```

jku (commented):
This makes sense to me.

SHARDING.md
haydentherapper: Updated with a curl-based example rather than cosign.

haydentherapper:
This is ready to go. It's been updated with example PRs based on when @cmurphy and I turned up the first shard. Everything under `## Steps` and `## Verify Shard Health` should be complete and up-to-date.

There's two WIPs:

* Probers. Shard discovery should be automatic, both for the write path with the SigningConfig and read path with the TrustedRoot. I added a TODO to the prober tracking issue to allow for us to manually set the shard URL in a GitHub Actions workflow so that we can run tests against a new shard before distributing the updated TUF metadata.
* Turn down/freezing. We haven't gone through this process yet, so I'm sure there will be bumps. In particular, we need to figure out the order of operations for tearing down NEGs. We've already added support in Terraform for deleting the signing key and databases.

I also assume we'll update the TUF playbook part after we go through it with staging.

SHARDING.md
```diff
@@ -0,0 +1,451 @@
...
+test the write path for the new shard. This will sign an entry with a private key and upload
+the signing event to the log.
+
+## WIP: Update Prober
+
+Tracked in https://github.com/sigstore/rekor-tiles/issues/46. We're aiming to avoid
+needing to manually update the prober. The prober should pick up new shards based on the
+SigningConfig (for the write path) and TrustedRoot (for the read path).
+
+We will need a mechanism for running prober tests against a new deployment before we've
```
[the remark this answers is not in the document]
ramonpetgrave64: Also, v2 shards aren't directly discoverable from the trusted root, because the trusted root does not contain `apiMajorVersion`, like the signing config does. We can work around this by querying `<shard url>/api/v2/checkkpoint`, though.

SHARDING.md
```diff
@@ -0,0 +1,451 @@
...
+### Post on Slack
+
+Make a post on Slack letting the community know we'll be freezing
+and turning down the log.
+
+### Disable any alerts
+
+Make sure to either disable or remove via Terraform any alerts
+for failing healthchecks or missing metrics. As we turn down the service
+and delete resources, we expect that healthchecks will start to fail.
```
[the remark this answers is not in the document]
haydentherapper: Changed this section to a WIP since we don't have the list of alerts yet.

SHARDING.md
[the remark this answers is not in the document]
haydentherapper: I updated this to specify that you should use the SigningConfig for shard discovery, as it will contain both the active and all inactive shards. If a prober needs to correlate between the SigningConfig and TrustedRoot, it can use the shard URL.

SHARDING.md
```diff
@@ -0,0 +1,451 @@
...
+which point you'd have a complete Sigstore stack on GCP. The downside of this approach is
+private deployers that don't need all GCP infrastructure have to initialize all resources,
+though deployers can always initialize each module referenced in `sigstore` individually.
+
+The other downside to referencing the `sigstore` module is that an update to a sub-module
+will concurrently update all deployed shards.  Declaring each shard as its own module allows us
+to make updates to shards one-by-one.
+
+The tiled transparency log module we'll reference in Terraform is
+[`tiles_tlog`](https://github.com/sigstore/terraform-modules/tree/main/gcp/modules/tiles_tlog).
```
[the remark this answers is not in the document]
haydentherapper: This is just to give background, either seems fine.

SHARDING.md
```diff
@@ -0,0 +1,465 @@
...
+and new TUF repository is published. Overestimate (by an hour or even a day) if needed,
+as you don't want clients to start using the new shard before all clients
+have picked up the latest TrustedRoot.
+* Previous shard's end for TrustedRoot: Clients will consider this shard's entry timestamps invalid
+if they are after this date. This must be after the start of the new shard. It doesn't have to
+overly precise, and can be updated in a future TUF signing event.
+
+### Retrieve shard public key
+
+Look at the service logs to find the public key. This is the easiest way to retrieve
```
ramonpetgrave64, line 307: Add a link to the service logs for an existing shard, so we can easily know how to find the logs for the next shard.

haydentherapper: The audience for this doc is someone familiar with the infrastructure so they should be familiar with where to retrieve logs for pods.

SHARDING.md
```diff
@@ -0,0 +1,464 @@
...
+
+Make sure to either disable or remove via Terraform alerts for the old shard
+for failing healthchecks for the write path or missing metrics. As we turn down the service
+and delete resources, we expect that healthchecks will start to fail.
+
+As we create alerting, we'll finalize this section with what alerts need to be removed.
+
+### Freeze the log
+
+Pass `--read-only=true` to the service, which will stop write traffic and
```
cmurphy, line 424: I think we could skip this step and just turn off the K8s deployment.

SHARDING.md
```diff
@@ -0,0 +1,464 @@
...
+2. Run `terraform plan` and `terraform apply` using the GHA workflow.
+
+3. [Install the gcloud CLI](https://cloud.google.com/sdk/docs/install) if not already installed.
+
+4. Authenticate to gcloud with application credentials: `gcloud auth application-default login`
+
+5. Using [tinkey](https://developers.google.com/tink/tinkey-overview#installation), create an encrypted Tink keyset:
+
+```
+tinkey create-keyset --key-template ED25519 --out enc-keyset.cfg --master-key-uri gcp-kms://projects/<project name>/locations/<region>/keyRings/<key-ring>/cryptoKeys/checkpoint-signer-key-encryption-key
```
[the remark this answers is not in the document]
haydentherapper: For the 2025-1 shard, [the key name variable](https://github.com/sigstore/public-good-instance/blob/3c97f5b6b6f52d2eb38018619555c57026241a67/terraform/environments/staging/1-infrastructure/staging.tf#L242) is set to `checkpoint-signer`. For future shards, we can just omit the key name variable and the use the new default `checkpoint-signer-key-encryption-key`.

SHARDING.md
```diff
@@ -0,0 +1,464 @@
...
+exclusively in memory.
+
+To create a shard KEK:
+
+1. Grant yourself access to use the new KEK. Update the `staging.tf`
+or `production.tf` config to include:
+
+```
+resource "google_kms_key_ring_iam_member" "rekor-tiles-keyring" {
+  key_ring_id = "<project-rekor or projectsigstore-staging>/<location, e.g. global>/<name of shard keyring>"
```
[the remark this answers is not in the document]
haydentherapper: Thanks, I added how the string is constructed.

SHARDING.md
```diff
@@ -0,0 +1,464 @@
...
+Merge and wait for ArgoCD to shut down the server.
+
+### Delete GCP resources
+
+We only need a subset of GCP resources to serve read traffic. To save costs,
+we can turn down the databases and destroy the KEK (indirectly destroying the signing key).
+
+First, remove the protection bit on the databases. In the Terraform configuration,
+set `spanner_database_sequencer_deletion_protection` and `spanner_database_antispam_deletion_protection`
+to `true`. Create a PR, merge, `plan`, and `apply`.
```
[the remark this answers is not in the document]
haydentherapper: Whoops, thanks.

SHARDING.md
haydentherapper: I'm good with that. The only benefit of freezing the log is to gracefully reject write traffic, but unless we're adding in a waiting period for clients to fail and update to the latest shard (which shouldn't be an issue as clients should be using TUF), this doesn't add much.

SHARDING.md
```diff
@@ -0,0 +1,464 @@
...
+can verify proofs from the shard and use the shard during signing.
+
+For TrustedRoot, the validity window for the shard should be from the point
+when the shard was spun up. For SigningConfig, the validity window should start
+after the TUF timestamp expires to guarantee all clients will have picked up
+the latest TrustedRoot. In practice for our TUF root with timestamp validity
+of one week, this means setting the validity window to be at least one week
+past when the TUF targets are signed.
+
+You will need two timestamps:
```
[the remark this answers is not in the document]
haydentherapper: Thanks, updated.

cmurphy (approved):

haydentherapper:
Will merge and iterate on this after we practice this playbook.

PR #255, opened 2025-04-23
Document required client changes for Rekor v2

### LINKED ISSUE #108: Publish doc summarizing necessary client changes
**Description**

Wrote up a summary in https://sigstore.slack.com/archives/C03SZ6SHU90/p1742236064115549?thread_ts=1742235329.158429&cid=C03SZ6SHU90, copied below. We should flesh this out into a brief doc for clients.

> For updating to the new upload API, it's relatively straightforward. It's a new request object defined via proto, and clients can use gRPC or HTTP/JSON. We support only DSSE and hashedrekord types. The response is the protobuf-specs defined TransparencyLogEntry, so clients no longer need to mutate Rekor's response into the bundle-compatible struct.
> 
> One change in terms of SLO is that Rekor will batch requests now (1-5s, TBD), so that it can return a checkpoint that eventually will also be co-signed by witnesses. This means clients will have to wait for a response and update timeouts.
> 
> The read APIs are significantly changed, but that shouldn't affect the core clients, only monitors, which have to download tiles rather than individual entries. Offline verification should be done by all clients already - if not, a client needs to change to support offline verification.
> 
> We've dropped SignedEntryTimestamps, so clients must fetch an RFC3161 timestamp. We'll be spinning up a public-good TSA, and clients can use their own preferred TSA as well.
> 
> For verifying proofs, it should be the same - verify an inclusion proof and checkpoint signature. Once we add witness co-signatures, clients may need to update how they parse the checkpoint, since co-sigs are additional lines on the signed note, and I think most clients expect only one signature. Witnessing will have its own design doc soon.
> 
> The URL for Rekor v2 and future shards will be published via the SigningConfig, so clients must support the config going forward. We recently updated the SigningConfig to support per-service API versions (so that clients can fetch either the Rekor v1 or v2 endpoint depending on support) and validity windows (for sharding, to prevent a client from writing to a log before the verification key is fully distributed).
> 
> We'll update the conformance test suite to verify clients support Rekor v2's API, RFC3161 timestamps, offline proof verification, and SigningConfig changes.

haydentherapper:
Fixes #108


CLIENTS.md
```diff
@@ -0,0 +1,264 @@
...
+Rekor will no longer return SignedEntryTimestamps or include integrated time
+in the response. Clients must fetch an RFC 3161 signed timestamp from a trusted
+timestamp authority and include the signed timestamp in the bundle.
+
+Sigstore now operates a timestamp authority at `timestamp.sigstore.dev` and
+`timestamp.sigstage.dev` for staging, and the roots for these services will
+be included in the TrustedRoot distributed via TUF. Clients may request timestamps
+from other trusted timestamp authorities as well.
+
+## SigningConfig support
```
haydentherapper, line 91: @jku I added some more info about the SigningConfig and motivation behind the changes, lemme know if you have more questions that I can clarify in this doc

CLIENTS.md
```diff
@@ -0,0 +1,264 @@
...
+## Signed RFC 3161 Timestamps
+
+Rekor will no longer return SignedEntryTimestamps or include integrated time
+in the response. Clients must fetch an RFC 3161 signed timestamp from a trusted
+timestamp authority and include the signed timestamp in the bundle.
+
+Sigstore now operates a timestamp authority at `timestamp.sigstore.dev` and
+`timestamp.sigstage.dev` for staging, and the roots for these services will
+be included in the TrustedRoot distributed via TUF. Clients may request timestamps
+from other trusted timestamp authorities as well.
```
loosebazooka, line 89: I assume we are expecting clients that use other "trusted timestamp" authorities to inject that information into the trusted root? Client code should not fetch the timestamp cert from the provider

CLIENTS.md
```diff
@@ -0,0 +1,266 @@
...
+## Rekor v2 API
+
+Rekor v2 supports HTTP or gRPC, like Fulcio. For Go, we have implemented a client
+already. For other languages, they can either use the OpenAPI docs or generated
+gRPC service proto, or create their own client.
+
+The service implements one write API, `/api/v2/log/entries`. Example JSON request bodies
+below:
+
+```json
```
loosebazooka, line 15: I haven't tested this, but `jsonc` appears to be a markdown rendering formation for json with comments.

CLIENTS.md
```diff
@@ -0,0 +1,266 @@
+# Client Changes for Rekor v2
+
+This document outlines the changes clients need to make to support
+Rekor v2.
+
+## Rekor v2 API
+
+Rekor v2 supports HTTP or gRPC, like Fulcio. For Go, we have implemented a client
+already. For other languages, they can either use the OpenAPI docs or generated
+gRPC service proto, or create their own client.
```
loosebazooka, line 10:
```suggestion
already. For other languages, they can either use the [OpenAPI docs](https://github.com/sigstore/rekor-tiles/tree/main/docs/openapi), [gRPC service proto](https://github.com/sigstore/rekor-tiles/tree/main/api/proto), or create their own client.
```

CLIENTS.md
```diff
@@ -0,0 +1,266 @@
...
+* Key IDs calculated per the [signed note spec](https://github.com/C2SP/C2SP/blob/main/signed-note.md#signatures)
+  * ECDSA and Ed25519 key IDs are based on the spec
+  * For RSA, the signature type is `0xFF` and we append `PKIX-RSA-PKCS#1v1.5`,
+    `key ID = SHA-256(key name || 0x0A || 0xFF || PKIX-RSA-PKCS#1v1.5 || public key)[:4]`
+
+## TrustedRoot With Multiple Logs
+
+Clients must verify that verification works with a TrustedRoot
+with multiple logs with overlapping validity windows. Confirm that
+the client will fetch the correct verification key from the TrustedRoot.
```
loosebazooka, line 220: Fetch correct key by matching logid/validity window?

CLIENTS.md
```diff
@@ -0,0 +1,266 @@
...
+
+// Request with a self-managed key
+{
+    "hashedRekordRequestV0_0_2": {
+        "digest": "<base64 digest of artifact>",
+        "signature": {
+            "content": "<base64 signature>",
+            "verifier": {
+                "publicKey": {
+                    "raw_bytes": "<base64 DER-encoded public key>"
```
cmurphy, line 38:
```suggestion
                    "rawBytes": "<base64 DER-encoded public key>"
```

CLIENTS.md
```diff
@@ -0,0 +1,266 @@
...
+            "payloadType": "<type, e.g. application/vnd.in-toto+json>",
+            "signatures": [
+                {
+                    "sig": "<base64-encoded signature>",
+                    "keyid": ""
+                }
+            ]
+        },
+        "verifier": {
+            "x509Certificate": "<base64 DER-encoded certificate>", # May also be a public key
```
cmurphy, line 60: Would be clearer if the example was repeated like you are for hashedrekord, with the `publicKey` field instead of `x509Certificate`

CLIENTS.md
```diff
@@ -0,0 +1,266 @@
...
+## Monitoring/Auditing
+
+One of the more significant changes in a tile-backed log is the changes
+to the read API. Inclusion and consistency proofs are not served via an
+API, rather the client requests the set of tiles necessary to compute
+the inclusion or consistency proof.
+
+When monitoring the log searching for entries, the monitor will not request
+entries by index, but by tile. Monitors can choose to only fetch complete
+tiles or request partial tiles. It is recommended to request partial tiles,
```
cmurphy, line 231: It might not be clear to someone brand-new to this what a partial tile is or why you would want it. Maybe add some clarification that the latest tile is generally a partial tile.

CLIENTS.md
haydentherapper: Added. I'd say we use log ID for now. We should later align log ID with how we're computing log IDs for checkpoints.

haydentherapper:
Proposing we change how clients find the correct tlog to use to verify a proof, in https://github.com/sigstore/protobuf-specs/pull/629. Will wait for reviews before updating this doc.

jku (commented):
I think the one case where I'm not confident is where the v2 rekor instance is added to signing config:
* before any clients are updated this is straight forward: clients keep signing with v1, nothing breaks
* when the client A is upgraded to support v2, the docs seem to imply that it should then only use the v2 instance when the signingconfig includes one  (correct?)
* this leads to a client incompatibility if other clients are not able to verify rekor v2 at this point yet

Basically we may need to confirm verification support across the ecosystem before we make signingconfig changes where a rekor v2 instance is enabled

CLIENTS.md
```diff
@@ -0,0 +1,266 @@
...
+message to store in the bundle.
+
+## Handling Longer Requests
+
+Clients need to increase request timeouts to at least 10 seconds.
+
+Rekor now batches uploads so that checkpoints are published less frequently.
+Additionally, we plan to support synchronous witnessing, where third-party
+witnesses independently verify the consistency of the log and Rekor provides
+co-signed checkpoints with each upload response.
```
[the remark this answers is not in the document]
haydentherapper: I'll add the line about uploading in parallel, but the client shouldn't assume the same checkpoint for each, as Tessera might trigger a new checkpoint creation in the middle of the set of uploads. If a client needs to save space, it would need to request the latest checkpoint and compute inclusion proofs for all entries, but once we introduce synchronous witnessing, this might be an issue unless witnesses also add support for requesting previously co-signed checkpoints.

haydentherapper:
> I think the one case where I'm not confident is where the v2 rekor instance is added to signing config:
> 
> * before any clients are updated this is straight forward: clients keep signing with v1, nothing breaks
> * when client A is upgraded to support v2, the docs seem to imply that it should then only use the v2 instance when the signingconfig includes one  (correct?)
> * this leads to a client incompatibility if other clients are not able to verify rekor v2 at this point yet
> 
> Basically we may need to confirm verification support across the ecosystem before we make signingconfig changes where a rekor v2 instance is enabled

cc @loosebazooka for something to discuss at the client meeting

This is a consequence of making a TLE v2 struct unfortunately, though even if we didn't, not all clients support signed timestamps so a TLE without an integrated timestamp wouldn't be verifiable anyways - there would have been incompatibility regardless. There will be a period in which some clients (-ruby and -js) will not be able to verify bundles produced by other clients.

I'd like to chat more with GitHub to understand how much cross-compatibility we need between -js and -go for artifact attestations, if any is actually needed.

CLIENTS.md
```diff
@@ -0,0 +1,369 @@
...
+
+Log ID, the SHA-256 digest of the log's public key, is used as a "unique"
+identifier for a log. The TransparencyLogEntry message includes it in a
+bundle, and clients use that log ID to lookup the correct log public key
+in the TrustedRoot to verify the bundle.
+
+Rekor will no longer include log IDs in the response, and instead clients
+should use the checkpoint key ID as specified in the
+[C2SP spec](https://github.com/C2SP/C2SP/blob/main/signed-note.md#signatures)
+to lookup the correct log public key to verify a bundle. Each transparency
```
jku, line 138: I think this might not be clear enough like this. How exactly do clients calculate the checkpoint key id for an entry -- is entrys checkpoint origin the "key name"?
 
 is there an example implementation of this?

CLIENTS.md
```diff
@@ -0,0 +1,369 @@
...
+            "keyDetails": "PKIX_ECDSA_P256_SHA_256"
+        }
+    }
+}
+```
+
+The response will be a
+[`TransparencyLogEntry` message](https://github.com/sigstore/protobuf-specs/blob/5296f13d62e7fad428581d969f664c30cc52f549/protos/sigstore_rekor.proto#L94),
+which should be persisted in a bundle. Clients no longer need to transform the
+Rekor response into a `TLE` message to store in the bundle.
```
ramonpetgrave64, line 98: It would be helpful to show the shapes of the old and new responses.

ramonpetgrave64 (changes requested):

haydentherapper:
Note to self: Split into changes for signingconfig, trustroot, and bundle. Add KindVersion and log ID (which is not hex-encoded, unlike rekor v1)

CLIENTS.md
```diff
@@ -0,0 +1,369 @@
...
+Rekor v2 only supports signature verification using a certificate or a
+public key, dropping support for PGP, minisign, pkcs7, SSH and TUF.
+Additional verifiers may be added in the future, but this will also require
+updating the client specification.
+
+### Handling Longer Requests
+
+Clients need to increase request timeouts to at least 10 seconds.
+
+Rekor now batches uploads so that checkpoints are published less frequently.
```
cmurphy, line 120: Not that this matters to clients but this is technically not true, the checkpoint publishing interval is not related to the batch interval.

CLIENTS.md
```diff
@@ -0,0 +1,369 @@
...
+
+Rekor will no longer include log IDs in the response, and instead clients
+should use the checkpoint key ID as specified in the
+[C2SP spec](https://github.com/C2SP/C2SP/blob/main/signed-note.md#signatures)
+to lookup the correct log public key to verify a bundle. Each transparency
+log instance in the TrustedRoot will include a `checkpoint_key_id` instead
+of a `log_id`.
+
+For more information, log IDs are not necessarily unique identifiers for
+a log, since a log may reuse its public key among instances. Additionally,
```
cmurphy, line 143: Why would we not just create a new key for the new instance?

CLIENTS.md
```diff
@@ -0,0 +1,369 @@
...
+the log origin is not necessarily a unique identifier, because multiple logs
+may be hosted by one origin. Even the combination of both is not necessarily
```
cmurphy, line 145: Each shard should be uniquely named

CLIENTS.md
```diff
@@ -0,0 +1,369 @@
...
+            },
+        }
+    ],
+
+    // Rekor service URLs. This example shows multiple active logs.
+    // The client should select the log with the highest API version
+    // it supports.
+    "rekorTlogUrls": [
+        {
+            "url": "https://rekor-2025-1.sigstore.dev",
```
cmurphy, line 224:
```suggestion
            "url": "https://log2025-01.rekor.sigstore.dev",
```

cmurphy, line 224:
```suggestion
            "url": "https://log2025-01.rekor.sigstore.dev",
```

CLIENTS.md
haydentherapper: Thanks, added a mention that this SHOULD NOT happen. Technically, a deployment can reuse the key and/or origin, which the checkpoint key ID calculation takes into account.

CLIENTS.md
haydentherapper: Added a line that for the public deployment, we always will. Private deployments may reuse keys.

CLIENTS.md
haydentherapper: Thanks, removed the mention of batching and clarified that this is just about blocking until a checkpoint has been published.

CLIENTS.md
haydentherapper: Added selection algorithms for this, thanks. The `TransparencyLogInstance` proto has been updated as well to document this.

tl;dr either 1) compare the bundle's `log_id` to the trust root log instance's `checkpoint_key_id` or `log_id` (what most clients are doing already), 2) enumerate all log instances and compare the checkpoint's key ID from the signature line to the first 4 bytes of the trust root log instance's `checkpoint_key_id`/`log_id` (which is what sigstore-python is doing), or 3) compute the log instance's checkpoint key ID using the instance's public key and origin and compare (this is Hard Mode, and I doubt any client will)

CLIENTS.md
```diff
@@ -367,3 +367,59 @@ be distributed by Sigstore's TUF root.
...
+  "inclusionProof": {
+    "logIndex": "0",
+    "rootHash": "OWI5MjU1MWIxZTA0NGIzNmQ0ZGE0MTU0M2UwNjEwZjVhNzQyMDNiN2JiOWEzOWYwNDExOTZlMTI2MTg5NmM3NQ==",
+    "treeSize": "1",
+    "hashes": [],
+    "checkpoint": {
+      "envelope": "rekor-local\n1\nm5JVGx4ESzbU2kFUPgYQ9adCA7e7mjnwQRluEmGJbHU=\n\n— rekor-local 2AtEIIwnbtxrneJ7L1lQebfBRl7TxK84DTmx+kcZi7A25cBDgESI23f9ylThAlOireJ7U+H8eZF/4kJQcn9o5Qt8mQU=\n"
+    }
+  },
+  "canonicalizedBody": "eyJhcGlWZXJzaW9uIjoiMC4wLjIiLCJraW5kIjoiaGFzaGVkcmVrb3JkIiwic3BlYyI6eyJoYXNoZWRSZWtvcmRWMF8wXzIiOnsiZGF0YSI6eyJhbGdvcml0aG0iOiJTSEEyXzI1NiIsImRpZ2VzdCI6ImR5ajRlZG5ZSGpONC96c2pqQmVlTGFoUzlzbHA5N1o2N0xUQVZ4anJqWHc9In0sInNpZ25hdHVyZSI6eyJjb250ZW50IjoiTUVRQ0lCK1lQYTlvM1NOMHNRNHVkdUdmK21aeHdGZk9oRlowQ2d5K3A3VnQxbzJTQWlBUEZESHFPQUpMWW12dENXT3NEeU5ZMUg0VjN6bTRORURZczNOeXZIaDFQZz09IiwidmVyaWZpZXIiOnsia2V5RGV0YWlscyI6IlBLSVhfRUNEU0FfUDI1Nl9TSEFfMjU2IiwicHVibGljS2V5Ijp7InJhd0J5dGVzIjoiTUZrd0V3WUhLb1pJemowQ0FRWUlLb1pJemowREFRY0RRZ0FFMnNsT2Y4ZVpjajJtb1cydDRVRmo3dkNMNlFwRHprRHFxU1VtbTRPSkNWdklhdUtMeG0wYUdzM1ZNUFBmYXVNUGFNdXRuMC9zM2pnMHJyb0Z4b2ljeWc9PSJ9fX19fX0="
```
ramonpetgrave64, line 423: Also show the decoded canonicalized body.

haydentherapper: There is now a section on parsing the canonicalized body.

haydentherapper:
Thanks everyone for comments! I've addressed all comments now. The primary changes are adding a section on entry parsing, some details about bundle-related changes, more info about log instance lookup in the trusted root, and a section on testing against a real instance using `curl`. The SigningConfig documentation has also been updated in the protobuf-specs repo, it's linked here.

jku (approved):
This looks great, no notes from me after reading through it twice. Thanks for polishing.

CLIENTS.md
```diff
@@ -0,0 +1,631 @@
...
+### Certificate and Public Key Verifiers
+
+Rekor v2 only supports signature verification using a certificate or a
+public key, dropping support for PGP, minisign, pkcs7, SSH and TUF.
+Additional verifiers may be added in the future, but this will also require
+updating the client specification.
+
+### Handling Longer Requests
+
+Clients need to increase request timeouts to at least 10 seconds.
```
ramonpetgrave64, line 196:
```suggestion
Clients need to increase request timeouts to at least 10 seconds when creating entries.
```

haydentherapper: Added, and I've doubled the timeout recommendation to 20 seconds.

jku (approved):

PR #256, opened 2025-04-24
Bump sigstore/sigstore, update Tink

haydentherapper:
KeyHandleToSigner dropped the return hash function, so we now use the default SignerVerifier loader. This will also prevent the server from loading in a key with an unsupported signing-hash algorithm pairining.


loosebazooka (dismissed):
so it looks like the hash function is now embedded in the singer's publickey algorithm details or something?

`LoadDefaultSignerVerifier` sounds like an odd name for something that takes a parameter and can return one of any number of implementations based on that parameter.

haydentherapper:
Tink supports only ECDSA and Ed25519. There are standard pairings for the hash algorithms used for these, which is how `LoadDefaultSignerVerifier` knows what hash alg to use given a key.

Technically, Tink supports some other pairings, but we're not using Tink to do the signing, we're just using Tink to securely store the key. Using `LoadDefaultSignerVerifier` will make sure we'll sign checkpoints with a key-hash pairing that's supported by the ecosystem.

jku (approved):

jku:
sorry, I merged in a dumb order: this now requires an update

jku (approved):

PR #270, opened 2025-04-30
Implement APIs for read server

haydentherapper:
This allows for deployers to serve read traffic without a separate
CDN/proxy. This is discouraged for GCP, since it's more costly to serve
read traffic in terms of egress costs, but can be used for other
backends and for local testing.

Rebased on #269.

### DIFF
--- a/cmd/rekor-server/app/serve.go
+++ b/cmd/rekor-server/app/serve.go
@@ -34,6 +34,7 @@ import (
 	"github.com/sigstore/rekor-tiles/pkg/signerverifier"
 	"github.com/sigstore/rekor-tiles/pkg/tessera"
 	"github.com/sigstore/sigstore/pkg/signature"
+	tesseraLib "github.com/transparency-dev/trillian-tessera"
 )
 
 var serveCmd = &cobra.Command{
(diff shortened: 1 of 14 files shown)


cmurphy (commented):
> This is discouraged for GCP

I think we would *never* want to use this for GCP. Instead of adding a --serve-read-paths flag, could this just be directly tied to the backend in use?

cmd/rekor-server/app/serve.go
```diff
@@ -112,7 +112,12 @@ var serveCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		rekorServer := server.NewServer(tesseraStorage, readOnly, algorithmRegistry)
+		var rekorServer server.RekorServer
+		if viper.GetBool("serve-read-paths") {
+			rekorServer = server.NewReadServer(tesseraStorage, readOnly, algorithmRegistry)
```
haydentherapper, line 117: Need to fix - This breaks with `--read-only` because `tesseraStorage` is nil but is still needed on the read path.

haydentherapper:
> I think we would never want to use this for GCP. Instead of adding a --serve-read-paths flag, could this just be directly tied to the backend in use?

For POSIX, it depends. I'd imagine a common deployment would be to serve the read path via nginx or some other load balancer that serves the /tiles directory. If a deployer wouldn't want to stand up an additional load balancer and doesn't expect significant read traffic, just the server should be able to handle both read and write traffic (pending load tests to verify this).

Technically, if someone didn't want to deal with managing a load balancer or public buckets, they could serve read traffic for GCP from the service.

I wouldn't be opposed to only allowing a subset of supported backends to serve read traffic though.

PR #286, opened 2025-05-05
Include KindVersion in TLE response

haydentherapper:
This adds a missing required field for the kind and version of the entry, so that clients will know how to parse the canonicalized body.

This is a much simpler approach to the type factory in Rekor. At this point, given there are only two types, I don't think we should reimplement the factory and simply use case statements.

### DIFF
--- a/pkg/server/service.go
+++ b/pkg/server/service.go
@@ -25,6 +25,7 @@ import (
 	pbs "github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v1"
 	pb "github.com/sigstore/rekor-tiles/pkg/generated/protobuf"
 	"github.com/sigstore/rekor-tiles/pkg/tessera"
+	"github.com/sigstore/rekor-tiles/pkg/types"
 	"github.com/sigstore/rekor-tiles/pkg/types/dsse"
 	"github.com/sigstore/rekor-tiles/pkg/types/hashedrekord"
 	"github.com/sigstore/sigstore/pkg/signature"
@@ -122,6 +123,12 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
 		slog.Warn("failed to integrate entry", "error", err.Error())
 		return nil, status.Errorf(codes.Unknown, "failed to integrate entry")
 	}
+	kv, err := types.GetKindVersion(req)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid type, must be either hashedrekord or dsse")
+	}
+	tle.KindVersion = kv
+
 	_ = grpc.SetHeader(ctx, metadata.Pairs(httpStatusCodeHeader, "201"))
 	metricsCounter.Inc()
 	return tle, nil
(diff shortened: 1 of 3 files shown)


loosebazooka (approved):

cmurphy:
Could you also update switch statements like these https://github.com/sigstore/rekor-tiles/blob/364926072f047df0cd4a9c3285c52e2f7f2d5dc9/pkg/client/write/write.go#L121 to use this?

pkg/server/service.go
```diff
@@ -122,6 +123,12 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
 		slog.Warn("failed to integrate entry", "error", err.Error())
 		return nil, status.Errorf(codes.Unknown, "failed to integrate entry")
 	}
+	kv, err := types.GetKindVersion(req)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid type, must be either hashedrekord or dsse")
+	}
+	tle.KindVersion = kv
```
ramonpetgrave64, line 130: Add tests more tests so we can see the rendered responses

 - https://github.com/sigstore/rekor-tiles/blob/364926072f047df0cd4a9c3285c52e2f7f2d5dc9/pkg/server/service_test.go#L189C4-L189C10

haydentherapper:
Following up on comments in another PR with adding log ID.

loosebazooka:
oops, stamped this a little too hastily

pkg/server/service.go
ramonpetgrave64: I see that KindVersion is also missing from the CanonicalizedBody.

loosebazooka: That part is intentional. The type should already be known by then.

haydentherapper: Appu, how is the type derived for monitors reading each entry of the log? With a full bundle, a client will know how to parse it.

loosebazooka: I just realized the problem here.

ramonpetgrave64: @loosebazooka But isn't the canonicalized by supposed to have all of the values of TLE? That seems to be how it is in v1's responses.

loosebazooka: right, I gotta go look at this again.

haydentherapper: Thanks for pointing this out Ramon. We do need entry type information for clients that do not have a Sigstore bundle. I've proposed a fix in https://github.com/sigstore/rekor-tiles/pull/287, which wraps the entry in a one-of so that we retain kind/version information.

Another option Appu and I discussed was trying to replicate the Rekor v1 body with `apiversion`, `kind`, and `spec`. While this might let clients get further along in parsing the body, clients will eventually fail because they won't support the new v2 types.

Appu will be reopening the discussion about a TLE v2 message. Clients will need to update their verification path regardless.

haydentherapper: Also to clarify, `KindVersion` in the bundle is still needed if `CanonicalizedBody`, which is optional, is omitted and a client reconstructs the rekor entry from the artifact, signature and verifier. We don't need `KindVersion` explicitly stated if the message types are retained, which is what 287 proposes.

Issue #285, opened 2025-05-05
Complete returned TransparencyLogEntry


ramonpetgrave64:
**Description**

Currently the return `TransparencyLogEntry` is missing a few fields. Clients like sigstore-python will need these values to construct a Bundle.

* https://github.com/sigstore/rekor-tiles/blame/33ac78c21a34de8dc061f45deb9cd5eed9b00dfe/pkg/tessera/tessera.go#L166-L170
* https://pkg.go.dev/github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v1#TransparencyLogEntry

We will need to decide what are appropriate values for these fields. Here are some ideas.

1. **LogID**: Concatenation of the storage's `origin` and signer's `keyID`. Do we think that a key might be reused across multiple storages?
2. **KindVersion**: a [KindVersion](https://pkg.go.dev/github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v1#KindVersion), depending on if the request was a `hashrekord` or `dsse`, both v0.0.1
3. **IntegratedTime**: the current time

**Version**

rekor-tiles@7289406c57920a57150dea3f72a1d1d2c7407e18

jku:
https://github.com/sigstore/sigstore-python/issues/1364 is for handling checkpoint_key_id in sigstore-python

haydentherapper:
As discussed offline, we'll also add log ID to the response to avoid introducing more breaking changes on the verification path.

cc @loosebazooka

loosebazooka:
> As discussed offline, we'll also add log ID to the response to avoid introducing more breaking changes on the verification path.

I think we can deprecate it eventually, but no one has implemented an alternative yet.

PR #287, opened 2025-05-06
Include type and version of entry for client verification

haydentherapper:
We have decided to mirror the Rekor v1 response to include API version,
kind, and spec, and clients should use the former two to parse the spec.
Clients should be able to handle a kind and version that they don't
support and gracefully fail.

Note that kind and version is required in the entry in addition to the
bundle. The type information is used by monitors parsing entries from
the log, and within the bundle is used by clients that do not have the
canonicalized body.

Ref: https://github.com/sigstore/rekor-tiles/issues/285


haydentherapper:
This can be merged as-is, or we can wait until we make a decision on whether we'll have a new TLE v2 message. Either way, I don't think anything here will change with a TLE v2 message, as we will still need type information.

loosebazooka (approved):
Yeah this new Entry should have all the information to programmatically determine the type. I imagine its a bit easier for users of the proto libraries, the pure json implementation is a bit weirder looking I imagine

ramonpetgrave64:
Did you mean to add the KindVersion to the canonicalized body, as well?

haydentherapper:
Decision from offline discussion: Will be updating this to https://github.com/sigstore/rekor-tiles/pull/287#issuecomment-2859930858 so that we can continue to use TLE without creating TLE v2.

api/proto/entry.proto
```diff
@@ -0,0 +1,52 @@
...
+    ];
+    Spec spec = 3 [(google.api.field_behavior) = REQUIRED];
+}
+
+// Spec contains one of the Rekor entry types.
+message Spec {
+    oneof spec {
+        HashedRekordLogEntryV0_0_2 hashed_rekord_v0_0_2 = 1 [
+            (google.api.field_behavior) = REQUIRED,
+            json_name = "hashedRekordLogEntryV0_0_2"
```
loosebazooka, line 45: I think json name can just be hashedRekordV0_0_2 (and dsseV0_0_2)

loosebazooka:
Lgtm from my phone. Will take another look though when I get to a real computer

api/proto/entry.proto
haydentherapper: Thanks, updated

loosebazooka (approved):
I think there's space somewhere to test that we truly are "canonicalizing" the entry. But this looks fine for now.

PR #288, opened 2025-05-06
Use Go cross-compiler rather than emulated build platform

### LINKED ISSUE #229: Improve CI container release pipeline runtime
**Description**

Using QEMU, which is required for building linux/arm64 (useful for macOS) in addition to linux/amd64 on a single runner, causes a dramatic increase in build time from ~3 min to ~30 min. Building the images independently on runners that match the targeted architecture (there's an ubuntu-arm64 GHA runner) would vastly improve runtime, but we need a way to merge the images back together to have a multi-platform container.

I found a blog post, https://dev.to/mathbunnyru/how-we-made-jupyter-docker-stacks-multi-arch-3ic1, that walks through how we could do this.

haydentherapper:
Fixes #229

Following this example:
https://docs.docker.com/build/building/multi-platform/#cross-compiling-a-go-application

Rather than use QEMU to launch an ARM build environment, which is incredibly slow on GitHub Actions, this uses Go's native cross compilation to build for multiple platforms. The builder is pinned to the build platform, Linux, while the target platform is passed to the Go builder as an OS and architecture.

Ko wasn't used because it doesn't build a multi-platform image, it builds an image per platform.

Also bumped Go to the latest version.

### DIFF
--- a/.github/workflows/build_container.yml
+++ b/.github/workflows/build_container.yml
@@ -67,9 +67,7 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
-      # Set up QEMU and the builder for multi-platform builds
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
+      # Set up builder for multi-platform builds
       - name: Set up Docker Buildx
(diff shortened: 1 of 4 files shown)


jku (approved):
:+1: This looks a lot more sensible than qemu.

Are TARGETOS and TARGETARCH set by `docker/build-push-action` based on the platforms input?

haydentherapper:
> 👍 This looks a lot more sensible than qemu.
> 
> Are TARGETOS and TARGETARCH set by `docker/build-push-action` based on the platforms input?

Yep, each platform provided to docker build is split into the OS and architecture

PR #291, opened 2025-05-08
Make the v2 log entry compatible with v1 entry formats

loosebazooka:
Re-organize the rekor v2 log entry to be compatible with the v1 types.

#### Summary
This allow clients to continue to address entries from v1 and v2 logs in the same way. An approach we can take for now to unblock client testing.


pkg/server/service.go
```diff
@@ -84,7 +84,16 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
 			slog.Warn("failed validating hashedrekord request", "error", err.Error())
 			return nil, status.Errorf(codes.InvalidArgument, "invalid hashedrekord request")
 		}
-		serialized, err = protojson.Marshal(entry)
+		hrEntry := &pb.Entry{
```
haydentherapper, line 87: I have a slight preference to make this change to `ToLogEntry` for each type, like https://github.com/sigstore/rekor-tiles/pull/287/files#diff-2a03147d283232db2861d00d3281b65d506efae367704cbd41d981e03dfaa6beL34, rather than have the service be responsible for knowing the Kind and ApiVersion

loosebazooka:
oops didn't realize #287 existed. Will close this for that.

pkg/server/service.go
```diff
@@ -97,7 +106,16 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
...
+			Spec: &pb.Spec{
+				Spec: &pb.Spec_DsseV0_0_2{
+					DsseV0_0_2: entry,
+				},
+			},
+		}
+		serialized, err = protojson.Marshal(dsseEntry)
 		if err != nil {
 			slog.Warn("failed marshaling dsse request", "error", err.Error())
 			return nil, status.Errorf(codes.InvalidArgument, "invalid dsse request")
```
haydentherapper, line 121: We should also update line 144 to reuse the kind and version from the Entry, we can remove `types.GetKindVersion` then.

haydentherapper:
Hah, do you want to add an update to the existing PR, or do you want me to change it?

loosebazooka:
> Hah, do you want to add an update to the existing PR, or do you want me to change it?

 go for it

pkg/server/service.go
loosebazooka: yeah that's definitely much nicer

PR #292, opened 2025-05-08
Add log ID to TransparencyLogEntry response, TLE testing in e2e test

### LINKED ISSUE #285: Complete returned TransparencyLogEntry

haydentherapper:
Fixes #285

To minimize the transition for clients between Rekor v1 and v2, we will
include the log ID in the response. This log ID is used by clients to
look up the correct instance in a trust root, though some clients ignore
it and build a pool of verification keys instead, and compare a
truncated log ID to the checkpoint signature key prefix.

I've chosen to make the log ID equal to the non-truncated checkpoint key
ID. For clients that treat the log ID as an opaque string when looking
up the instance in the trusted root, no changes should be needed.
Similarly, if a client uses the log ID to compare against the checkpoint
key ID, like what sigstore-python does, no changes are needed.

In the TransparencyLogInstance in the trusted root, both the log ID and
checkpoint key ID will be set to the same value. Longer-term, we'll
deprecate log ID. Even longer-term, ideally clients would enumerate all
keys and compute the checkpoint key IDs themselves, but for now, we'll
keep populating the IDs in the trusted root for convenience.

### DIFF
--- a/cmd/rekor-server/app/serve.go
+++ b/cmd/rekor-server/app/serve.go
@@ -30,10 +30,12 @@ import (
 	"sigs.k8s.io/release-utils/version"
 
 	"github.com/sigstore/rekor-tiles/pkg/algorithmregistry"
+	"github.com/sigstore/rekor-tiles/pkg/note"
 	"github.com/sigstore/rekor-tiles/pkg/server"
 	"github.com/sigstore/rekor-tiles/pkg/signerverifier"
 	"github.com/sigstore/rekor-tiles/pkg/tessera"
 	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/sigstore/sigstore/pkg/signature/options"
 )
 
 var serveCmd = &cobra.Command{
@@ -77,6 +79,20 @@ var serveCmd = &cobra.Command{
 			slog.Error(fmt.Sprintf("failed to initialize append options: %v", err))
 			os.Exit(1)
 		}
+		// Compute log ID for TransparencyLogEntry, to be used by clients to look up
+		// the correct instance in a trust root. Log ID is equivalent to the non-truncated
+		// hash of the public key and origin per the signed-note C2SP spec.
+		pubKey, err := signer.PublicKey(options.WithContext(ctx))
+		if err != nil {
+			slog.Error(fmt.Sprintf("failed to get public key: %v", err))
+			os.Exit(1)
+		}
+		_, logID, err := note.KeyHash(viper.GetString("hostname"), pubKey)
+		if err != nil {
+			slog.Error(fmt.Sprintf("failed to get log ID: %v", err))
+			os.Exit(1)
+		}
+
 		readOnly := viper.GetBool("read-only")
 		var tesseraStorage tessera.Storage
 		shutdownFn := func(_ context.Context) error { return nil }
@@ -112,7 +128,7 @@ var serveCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		rekorServer := server.NewServer(tesseraStorage, readOnly, algorithmRegistry)
+		rekorServer := server.NewServer(tesseraStorage, readOnly, algorithmRegistry, logID)
 
 		server.Serve(
 			ctx,
--- a/pkg/note/note.go
+++ b/pkg/note/note.go
@@ -92,68 +92,72 @@ func isValidName(name string) bool {
 	return name != "" && utf8.ValidString(name) && strings.IndexFunc(name, unicode.IsSpace) < 0 && !strings.Contains(name, "+")
 }
 
-// genConformantKeyHash generates a 4-byte key ID for typical (non-ECDSA) keys.
-func genConformantKeyHash(name string, sigType, key []byte) uint32 {
+// genConformantKeyHash generates a truncated (4-byte) and non-truncated
+// identifier for typical (non-ECDSA) keys.
+func genConformantKeyHash(name string, sigType, key []byte) (uint32, []byte) {
 	hash := sha256.New()
 	hash.Write([]byte(name))
 	hash.Write([]byte("\n"))
 	hash.Write(sigType)
 	hash.Write(key)
 	sum := hash.Sum(nil)
-	return binary.BigEndian.Uint32(sum)
+	return binary.BigEndian.Uint32(sum), sum
 }
 
 // ed25519KeyHash generates the 4-byte key ID for an Ed25519 public key.
 // Ed25519 keys are the only key type compatible with witnessing.
-func ed25519KeyHash(name string, key []byte) uint32 {
+func ed25519KeyHash(name string, key []byte) (uint32, []byte) {
 	return genConformantKeyHash(name, []byte{algEd25519}, key)
 }
 
 // ecdsaKeyHash generates the 4-byte key ID for an ECDSA public key.
 // ECDSA key IDs do not conform to the note standard for other key type
 // (see https://github.com/C2SP/C2SP/blob/8991f70ddf8a11de3a68d5a081e7be27e59d87c8/signed-note.md#signature-types).
-func ecdsaKeyHash(key *ecdsa.PublicKey) (uint32, error) {
+func ecdsaKeyHash(key *ecdsa.PublicKey) (uint32, []byte, error) {
 	marshaled, err := x509.MarshalPKIXPublicKey(key)
 	if err != nil {
-		return 0, fmt.Errorf("marshaling public key: %w", err)
+		return 0, nil, fmt.Errorf("marshaling public key: %w", err)
 	}
 	hash := sha256.Sum256(marshaled)
-	return binary.BigEndian.Uint32(hash[:]), nil
+	return binary.BigEndian.Uint32(hash[:]), hash[:], nil
 }
 
 // rsaKeyhash generates the 4-byte key ID for an RSA public key.
-func rsaKeyHash(name string, key *rsa.PublicKey) (uint32, error) {
+func rsaKeyHash(name string, key *rsa.PublicKey) (uint32, []byte, error) {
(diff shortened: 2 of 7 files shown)


cmd/rekor-server/app/serve.go
```diff
@@ -77,6 +80,20 @@ var serveCmd = &cobra.Command{
 			slog.Error(fmt.Sprintf("failed to initialize append options: %v", err))
 			os.Exit(1)
 		}
+		// compute log ID for TransparencyLogEntry, to be used by clients to look up
```
haydentherapper, line 83: I've left this PR in draft, because I'd like to replace this with the checkpoint key ID instead, which is a truly unique identifier. If any clients are not treating this log ID as an opaque string, then I'll leave this as-is.

ramonpetgrave64: sigstore-python currently uses it to know which key to use to verify the checkpoints. Did you mean to replace the `log_id` field in `TransparencyLogEntry` with `checkpoint_key_id`?

- https://github.com/sigstore/sigstore-python/blob/e80427ffb754de117d62727f25b52cbf8508e84e/sigstore/_internal/rekor/checkpoint.py#L174-L178

I was getting ready to remove that requirement, but if we decide to merge this PR as is, then this patch in sigstore-python won't be necesasry.

- https://github.com/sigstore/sigstore-python/compare/main...ramonpetgrave64:sigstore-python:optional-log-id

ramonpetgrave64 (dismissed):
Would be useful to see a test will a fully rendered reponse.

cmd/rekor-server/app/serve.go
haydentherapper: `checkpoint_key_id` is a field on a tlog instance in the trust root.

I was planning to continue to populate the log ID so that clients don't have to make any changes with how they look up the correct instance to verify a proof. Over time, we'll deprecate log ID and clients will use the key ID from a signature line on the checkpoint to look up the correct instance, which is what sigstore-python is doing from what you sent. sigstore-python ignores the `log_id` in the TLE I believe.

For sigstore-python, the client assumes that the log ID (sha-256 digest of an encoded public key) and checkpoint key ID (calculated per the c2sp spec) are computed in the same way. For Rekor v1, this was always true, because we weren't using the c2sp spec calculation, we just always hashed the public key.\
For Rekor v2, this will not be true (expect for ECDSA keys, whose key IDs are SHA digests per the spec).

So either:

* In the tlog instance in the trust root, we set `log_id` and `checkpoint_key_id` to the same value, and clients can choose which to look at (this is for sigstore-python). We set the `log_id` equal to the checkpoint key ID (this is for clients that treat the log_id as an opaque string and use the TLE to look up the correct instance).
* We push this PR out, and sigstore-python would have to be updated to either compute the checkpoint key ID or use a not-yet-populated `checkpoint_key_id` in the instance in the trust root.

I'd prefer the first solution. Your proposed PR wouldn't be necessary then, it would continue working as expected.

haydentherapper: If we went with the second approach, rather than try all keys, the client would either need to use [`checkpoint_key_id`](https://github.com/sigstore/protobuf-specs/blob/8b886ffc927028be9df23035bf9034d23f551db9/protos/sigstore_trustroot.proto#L77), or compute it following https://github.com/C2SP/C2SP/blob/main/signed-note.md#signatures. `checkpoint_key_id` would be populated for convenience because I think it's unnecessary for every client to know how to compute a key ID.

But all of this would be moot if we just treat the log ID and checkpoint key ID as the same value. So far, looking at -python, -java and -go, it looks like we can do so without issue.

haydentherapper: I've updated this to have the log ID be equal to the non-truncated checkpoint key ID. For clients that treat log ID as an opaque string for lookup in the trusted root, no changes should be needed. For clients that compare the log ID to checkpoint signature's key prefix, like in sigstore-python, there should also be no changes necessary, since the log ID is the non-truncated checkpoint signature key prefix.

haydentherapper:
Updated the e2e tests to be far more comprehensive, to compare the response's TLE.

cmd/rekor-server/app/serve.go
```diff
@@ -77,6 +79,20 @@ var serveCmd = &cobra.Command{
...
 		}
+		// Compute log ID for TransparencyLogEntry, to be used by clients to look up
+		// the correct instance in a trust root. Log ID is equivalent to the non-truncated
+		// hash of the public key and origin per the signed-note C2SP spec.
+		pubKey, err := signer.PublicKey(options.WithContext(ctx))
+		if err != nil {
+			slog.Error(fmt.Sprintf("failed to get public key: %v", err))
+			os.Exit(1)
+		}
+		_, logID, err := note.KeyHash(viper.GetString("hostname"), pubKey)
```
loosebazooka, line 90: I'm not sure we should be using hostname here? Isn't this service potentially hidden behind a load balancer? Is it doubly used to define what the infrastructure level hostname should be and the externally visible hostname when computing the logId?

haydentherapper: `hostname` is actually just the origin string, it's not used to serve traffic (those flags are `(http|grpc)-(address|port)`. 

If one isn't set, then the OS hostname is used - https://github.com/sigstore/rekor-tiles/blob/b64cdc986066e24d483c26784c71f62b3cc6f760/cmd/rekor-server/app/serve.go#L151-L156. For any public deployment, we'll always set the hostname.

We could update this to be `--origin` in another PR.

loosebazooka: Ah interesting. Cool.

haydentherapper:
Rebased and updated the e2e test to compare the non-hex-encoded root hash.

pkg/note/note_test.go
```diff
@@ -79,24 +80,28 @@ fycQ0e/nw1JkypGoL1j4cwM=
 func TestKeyHash(t *testing.T) {
 	origin := "testkey"
 	tests := []struct {
-		name          string
-		key           []byte
-		expectedKeyID uint32
+		name             string
+		key              []byte
+		expectedKeyID    uint32
+		expectedLogIDHex string
```
ramonpetgrave64, line 86: Since LogID is meant to be bytes, is it still supposed to be hex-encoded first? It will be encoded again to base64 during transmission.

 - https://github.com/sigstore/protobuf-specs/blob/a4c70fea6889694dfbceea11e0f9147335a95d79/protos/sigstore_common.proto#L144-L145

haydentherapper: This is just for tests to store the log ID as a string. The log ID is bytes and JSON marshalled to a base64-encoded string in the response.

ramonpetgrave64: Ok, It would be better to use bas64 in these examples.

haydentherapper: I've changed this to call `hexDecodeOrDie` rather than pass around the hex-encoded ID to avoid confusion. I'm leaving this as a hex string since we aren't testing anything with the bundle and this is typical for digest output.

ramonpetgrave64: Ok. I'm not blocking, but there's also `b64DecodeOrDie`.

ramonpetgrave64 (approved):

pkg/note/note.go
```diff
@@ -93,67 +93,70 @@ func isValidName(name string) bool {
 }
 
 // genConformantKeyHash generates a 4-byte key ID for typical (non-ECDSA) keys.
```
loosebazooka, line 95: update comment for return type? to match `KeyHash`. Although there are all internal, maybe not relevant.

pkg/note/note.go
```diff
@@ -93,67 +93,70 @@ func isValidName(name string) bool {
...
 	}
 	rsaAlg := append([]byte{algUndef}, []byte(rsaID)...)
-	return genConformantKeyHash(name, rsaAlg, marshaled), nil
+	id, hash := genConformantKeyHash(name, rsaAlg, marshaled)
+	return id, hash, nil
 }
 
-// keyHash generates a 4-byte identifier for a public key/origin
-func keyHash(origin string, key crypto.PublicKey) (uint32, error) {
+// KeyHash generates a truncated and non-truncated 4-byte identifier for a
```
loosebazooka, line 135:
```suggestion
// KeyHash generates a truncated (4-byte) and non-truncated identifier for a
```

loosebazooka (approved):

pkg/note/note.go
haydentherapper: This comment is accurate, this only returns a 4-byte key ID for a non-ECDSA key.

loosebazooka: it's returning (uint32, []byte) which is 4bytes+sum?

haydentherapper: Oh sorry, yep, misread this. Updating the comment

loosebazooka (approved):

PR #293, opened 2025-05-09
Fix cross-compilation for dlv binary for Docker

haydentherapper:
When the build and target platform are the same, go install installs the binary under GOPATH/bin. If they differ, then it is installed under GOPATH/bin/TARGETOS_TARGETARCH. I didn't notice this in my local testing, since the build and target platforms of course matched. In CI, this caused an issue when the build platform (linux_amd64) did not match the target platform (linux_arm64), and the binary was installed in an unexpected directory.

Note that go install doesn't let you select output location, and go build doesn't let you target specific versions. There are feature requests for both of these.

Also copied in the ENV variables into the dlvbuilder stage, since ARG and ENV steps are not copied into later stages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,12 +44,20 @@ CMD ["rekor-server", "serve"]
 FROM builder AS dlvbuilder
 ARG TARGETOS
 ARG TARGETARCH
+ENV APP_ROOT=/opt/app-root
+ENV GOPATH=$APP_ROOT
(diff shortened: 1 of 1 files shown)


haydentherapper:
Confirmed working: https://github.com/haydentherapper/rekor-tiles/actions/runs/14918601689/job/41909544195

jku (approved):

PR #294, opened 2025-05-09
Return 501 Method Not Implemented for read APIs

haydentherapper:
When we were deploying the service to staging, we noticed that the read APIs were returning 200 with an empty body for some reason. The root cause is due to how we handle returning an Unimplemented gRPC status code.

In order to return the non-retryable 405 Method Not Allowed when the server goes into a read-only mode, we added special handling when the gRPC status code is set to Unimplemented. Otherwise, it returns a 501 Not Implemented, which is not an accurate status for when the server is read-only.

When the status code is Unimplemented, we set the response code and body in the httpResponseModifier function based on internally-set gRPC headers. This worked correctly for the read-only response. However, for the 3 unimplemented read APIs, while we were setting a gRPC response code of Unimplemented, we weren't setting the internal headers. This meant that we skipped over where the HTTP response code/body is set, and so we returned a default 200 with an empty body for the read APIs.

This PR modifies the HTTP error code handler to look for a custom header rather than a specific gRPC error code, so that when the gRPC code is Unimplemented without a custom header, it returns 501.

### DIFF
--- a/pkg/server/http.go
+++ b/pkg/server/http.go
@@ -35,11 +35,9 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	pb "github.com/sigstore/rekor-tiles/pkg/generated/protobuf"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/health/grpc_health_v1"
-	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 )
 
@@ -88,7 +86,7 @@ func newHTTPProxy(ctx context.Context, config *HTTPConfig, grpcServer *grpcServe
 	}
 	mux := runtime.NewServeMux(
 		runtime.WithMarshalerOption(runtime.MIMEWildcard, &strictMarshaler),
-		runtime.WithErrorHandler(methodNotAllowedHandler),
+		runtime.WithErrorHandler(customHTTPErrorHandler),
 		runtime.WithForwardResponseOption(httpResponseModifier),
 		runtime.WithHealthzEndpoint(grpc_health_v1.NewHealthClient(cc)), // localhost:[port]/healthz
 	)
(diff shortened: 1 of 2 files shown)


haydentherapper:
FYI @bobcallaway, this was the root cause of the 200 from the load balancer.

bobcallaway:
to me, a 405 indicates a client made a request for an HTTP METHOD against a resource that is not allowed (e.g. a write `POST` was sent to `/a/b/c` when only `GET` was supported for that resource).

I'm also confused about the statement about the service being "read-only" - if the service _is read-only_, shouldn't we explicitly *support* these endpoints? I guess I read it as whether the rekor-service is directly serving the tiles/entries/checkpoint or not; if it is, it should always return the resources if they exist or a 404 if they don't; a 501 not implemented feels correct if the rekor-service _isn't_ serving the content, and client retries are a result of a misconfigured load-balancer/url map.

haydentherapper:
> to me, a 405 indicates a client made a request for an HTTP METHOD against a resource that is not allowed (e.g. a write POST was sent to /a/b/c when only GET was supported for that resource).

From the link about 501s:

> Servers are required to support [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET) and [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD), and therefore must not return 501 in response to requests with these methods. If the server does recognize the method, but intentionally does not allow it, the appropriate response is [405 Method Not Allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/405).

I agree the text of the 501 seems more accurate, but I was going off the recommendation from Mozilla.

> I'm also confused about the statement about the service being "read-only" - if the service is read-only, shouldn't we explicitly support these endpoints? 

Not with the GCP deployment, the read path is always served from GCS. There's another PR out in which I implement these read APIs, though we would only enable that for select backends that require the read path (e.g. MySQL), not GCP.

bobcallaway:
https://datatracker.ietf.org/doc/html/rfc7231#section-4.1 states: `All general-purpose servers MUST support the methods GET and HEAD. ` However Rekor isn’t a general purpose web server. 

My logic is that if we support other methods for `/api/v2/checkpoint` but not GET, a 405 would make sense. The server is either configured to support inline read paths or not, and if it’s configured to _not_ support them, I think 501 is the correct response code per https://datatracker.ietf.org/doc/html/rfc7231#section-6.6.2: `This is the appropriate response when the server does not recognize the request method and is not capable of supporting it for any resource.` if we’re not serving reads from rekor, there’s no public facing GET endpoint from the service.

haydentherapper:
That makes sense, I'll update this to return 501s for the read methods.

I'm reviewing whether it's appropriate to return 405 when the log is frozen for POSTing a new entry. The server does understand the method, and doesn't support POSTs for the target resource, so I think 405 is accurate. It's maybe a little strange because there is no other method that is supported, but none of the other return codes seem precise enough.

loosebazooka (approved):

PR #295, opened 2025-05-10
Fix root hash as bytes rather than hex-encoded

haydentherapper:
To produce strings in Rekor v1 responses, all hashes were hex-encoded. When the protobuf specs were created, rather than specify strings for hashes, we specified bytes. This fixes the root hash of the inclusion proof to be a byte array (and when marshalled to JSON, it will be base64-encoded).

This is another example of what should be updated with a TLE v2, dropping log index (specified in the bundle), root hash and tree size (both specified in the checkpoint). I'll update the client spec to recommend ignoring these fields.

### DIFF
--- a/pkg/tessera/tessera_test.go
+++ b/pkg/tessera/tessera_test.go
@@ -27,10 +27,14 @@ import (
 	tessera "github.com/transparency-dev/trillian-tessera"
 )
 
-var tileHash = "81bfc09c412c04da53a1b0ddb94dce48d6a24e9ea58987f7d45a04e8007fb3ca"
+var tileHashHex = "81bfc09c412c04da53a1b0ddb94dce48d6a24e9ea58987f7d45a04e8007fb3ca"
 
 func TestAdd(t *testing.T) {
 	ctx := context.Background()
+	tileHash, err := hex.DecodeString(tileHashHex)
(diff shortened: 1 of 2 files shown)


pkg/tessera/tessera_test.go
```diff
@@ -27,10 +27,14 @@ import (
 	tessera "github.com/transparency-dev/trillian-tessera"
 )
 
-var tileHash = "81bfc09c412c04da53a1b0ddb94dce48d6a24e9ea58987f7d45a04e8007fb3ca"
+var tileHashHex = "81bfc09c412c04da53a1b0ddb94dce48d6a24e9ea58987f7d45a04e8007fb3ca"
 
 func TestAdd(t *testing.T) {
 	ctx := context.Background()
+	tileHash, err := hex.DecodeString(tileHashHex)
```
loosebazooka, line 34: I like colleen's helpers `DecodeOrDie` for these kind of things.

loosebazooka (approved):

pkg/tessera/tessera_test.go
haydentherapper: Sorry, missed this, fixing in https://github.com/sigstore/rekor-tiles/pull/292

Issue #194, opened 2025-04-08
publish protos for clients to consume


loosebazooka:
**Description**

provide packaged versions of https://github.com/sigstore/rekor-tiles/tree/main/api/proto for clients to consume. Clients should not be hand building json.

haydentherapper:
Completed in protobuf-specs

PR #303, opened 2025-05-15
Add client generation for python

loosebazooka:
Client code is intentionally not checked in here, this should move to protobuf-specs when ready. Clients may chose to generate and copy the code to somewhere they can use it.

```
cd protoc-builder
make python
```

Ref: https://github.com/sigstore/rekor-tiles/issues/194

#### Summary
Adding make targets for generating python code


loosebazooka:
fyi @ramonpetgrave64  -- I think this should help, it doesn't generate a grpc client, but I think most people are using the http endpoints anyway

haydentherapper (commented):
Why is this needed for only python and not other libraries? Should we start moving the protos over to the specs repo to avoid duplication in the builder?

Regarding OpenAPI, did we figure out if the generated files are complete or not?

loosebazooka:
It's needed for everyone. This is to unblock the work for Python client. The others can be added. I'm working on fixing the open API defs. It turns out "Entry" isn't referenced by the service definition so its ignored. I'm looking into that config.

protoc-builder/Dockerfile.python
```diff
@@ -0,0 +1,33 @@
+A#
```
haydentherapper:
```suggestion
#
```

protoc-builder/Makefile
```diff
@@ -51,13 +54,27 @@ protos: proto-image
...
+	@echo "Generating python proto Docker image"
+	${DOCKER_BUILD} -t ${PROTOC_PYTHON_IMAGE} -f Dockerfile.python \
+    --build-arg PROTOC_BASE=${PROTOC_REKOR_IMAGE} \
+    .
+
+.PHONY: protoc-image
+protoc-image: Dockerfile
 	@echo "Generating go proto Docker image"
 	${DOCKER_BUILD} -t ${PROTOC_REKOR_IMAGE} -f Dockerfile \
-    --build-arg BUILDKIT_SYNTAX=docker/dockerfile:1 \
```
haydentherapper, line 58: What's the reason for dropping this? I'm not that familiar with Buildkit, but I think this improves build performance

haydentherapper (dismissed):
Confirmed `make python` works on darwin/amd_64

haydentherapper:
Once we have a client further along and have more confidence in the current API, let's get the generation moved out to protobuf-specs.

protoc-builder/Makefile
loosebazooka: It's a weird commandline override of the syntax. I think we actually need to set the syntax in a comment on each dockerfile: https://docs.docker.com/build/buildkit/frontend/

to docker/dockerfile:1

protoc-builder/Dockerfile.python
```diff
@@ -0,0 +1,35 @@
...
```
haydentherapper, line 1: A :)

protoc-builder/Makefile
haydentherapper: Good find, thanks

protoc-builder/Dockerfile.python
loosebazooka: oh I didn't rebase oops

haydentherapper (approved):

Issue #289, opened 2025-05-07
Client support for Rekor V2: sigstore-python


ramonpetgrave64:
**Description**

working dev branch: 
 - https://github.com/sigstore/sigstore-python/compare/main...ramonpetgrave64:sigstore-python:rekor-v2
   - https://github.com/sigstore/sigstore-python/pull/1387
 - only with hashedrekord request

- [x] Rekor v2 support (WIP, @ramonpetgrave64)
  - [x] Hashedrekord
  - [x] DSSE
  - [x] Verification with a certificate
  - [ ] Verification with a public key (won't be supported, isn't currently)
  - [x] Longer timeout https://github.com/sigstore/sigstore-python/issues/1448
- [x] RFC3161 support
- [x] SigningConfig (WIP, @jku)
- [x] Remove online verification/search
- [x] Checkpoint verification conforms to [C2SP spec](https://github.com/C2SP/C2SP/blob/main/tlog-checkpoint.md)
- [ ] Not relying on attestation storage
- [x] Compatibility fixes:
   - [x] dsse support
  - [x] https://github.com/sigstore/sigstore-python/pull/1373
  - [x] https://github.com/sigstore/sigstore-python/pull/1366
  - [x] https://github.com/sigstore/sigstore-python/pull/1368
  - [x] https://github.com/sigstore/sigstore-python/pull/1370
  - [x] https://github.com/sigstore/sigstore-python/pull/1377
  - [x] https://github.com/sigstore/sigstore-python/pull/1381
  - [x] fix improper decoding of the root hash (not hex)
    - dev branch: https://github.com/sigstore/sigstore-python/compare/main...ramonpetgrave64:sigstore-python:root-hash-decode
    - likely fixed in https://github.com/sigstore/rekor-tiles/pull/295
  - [x] https://github.com/sigstore/sigstore-python/pull/1382
  - [x] proper request crafting
  - [x] proper response parsing
    - [x] add hashedrekord v002 and dsse v002
  - [x] verification: confirm expected entry canonicalized body
- [ ] testing
  - [x] verify bundles from other languages -- manually tested some, should add more bundles to conformance as supporting clients appear
  - [x] using v2 at sigstage.dev
  - [ ] using the new setup-sigstore-env scaffolding https://github.com/sigstore/sigstore-python/pull/1441
  - [x] meanwhile trust-config is being prepared, hardcode compatibility for v2.
    - https://github.com/sigstore/sigstore-python/pull/1363
[...]

PR #304, opened 2025-05-15
fix pydantic generation

ramonpetgrave64:
https://github.com/sigstore/rekor-tiles/issues/289

Followup to #303 

#### Release Note

Fixes the option for specifying pydantic types.

 - https://github.com/danielgtaylor/python-betterproto/blob/124613f55fa8ff62a460e1ad33c5c78723b703a0/README.md#generating-pydantic-models

> With the important change being --python_betterproto_opt=pydantic_dataclasses. This will swap the dataclass implementation from the builtin python dataclass to the pydantic dataclass. You must have pydantic as a dependency in your project for this to work.

### Testing

The existing code doesn't produce an error, but here's a sample of the different results with the new change

```python
-from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from dataclasses import dataclass
+else:
+    from pydantic.dataclasses import dataclass
```

#### Documentation

Fix pydantic typing in the python code generator.


loosebazooka (approved):
Ah interesting, good catch

PR #305, opened 2025-05-16
Use 3.10 style for typing imports

ramonpetgrave64:
#### Summary

Uses python 3.10 style for some types, since `typing.List` is deprecated and will error with the ruff linters.

 - https://github.com/danielgtaylor/python-betterproto/blob/124613f55fa8ff62a460e1ad33c5c78723b703a0/README.md#310

#### Release Note

Use 3.10 style for typing imports.


ramonpetgrave64:
Not assuming that we do, just preparing for when we do. Even with specifying the python version, ruff will still error when importing `typing.List.` We could instruct ruff to ignore those errors now, but then we will want to remove that instruction after we upgrade to python3.10.

jku:
Is till don't get where this happens: does the issue with ruff just come up because you currently embed the generated sources in a sigstore-python?

Because we should probably not be running ruff on generated code.

PR #317, opened 2025-05-20
App: Log signing key public key on startup

### LINKED ISSUE #306: Log the public key on server startup
**Description**

The public key is needed for verifying checkpoint signatures. For the Sigstore public instance, we distribute the log public key with TUF.

For Rekor v1, we provided a public endpoint to fetch the log public key. This has been misused by clients that fetch the key and then verify the checkpoint - keys must be fetched out of band. For this reason, we removed the endpoint in v2.  

For deployments that don’t have easy access to the key material, like with a tink keyset (with a KMS key signer, you can just request the public key from the cloud KMS provider), it’d be useful to print the public key on startup.

I’d recommend adding a startup log statement that prints the PEM encoded public key.

jku:
The output is the base64 encoded DER (so PEM without headers).

Fixes #306

---

Output from docker compose up (the testing docker setup):
`rekor-1       | 2025/05/20 07:25:14 INFO Loaded signing key "pub key PEM"="MCowBQYDK2VwAyEAREvJyNZGjX6B3DAIuD3BTg9rIwV00GY8Xg5FU+IFDUQ="
`

This matches tests/testdata/pki/ed25519-pub-key.pem.


cmd/rekor-server/app/serve.go
```diff
@@ -74,6 +76,18 @@ var serveCmd = &cobra.Command{
...
+		if err != nil {
+			slog.Error(fmt.Sprintf("failed to get public key from signing key: %v", err.Error()))
+			os.Exit(1)
+		}
+		der, err := x509.MarshalPKIXPublicKey(pubkey)
+		if err != nil {
+			slog.Error(fmt.Sprintf("failed to marshal public key to DER: %v", err.Error()))
+			os.Exit(1)
+		}
+		slog.Info("Loaded signing key", "pub key PEM", base64.StdEncoding.EncodeToString(der))
```
loosebazooka, line 89: pub key PEM -> pub key base64 encoded DER ?

jku: My understanding is that  PEM is exactly that (surrounded by the header/footer string). 

I can look for a pem specific api to do that if it seems better though

loosebazooka: PEM requires the header and footer though?

loosebazooka: I think keeping it base64 encoded DER is fine, since thats what we do in trusted_root.json and stuff.

jku: Ah I see, you are saying "base64 encoded DER" would be a more useful description than "PEM".

I'm not sure I completely agree but I can change that.

haydentherapper: I love an excuse to look at an RFC :)

https://datatracker.ietf.org/doc/html/rfc7468#section-3, based on https://datatracker.ietf.org/doc/html/rfc1421, notes that PEM encodings begin and end with a header. 

I do agree with printing base64-encoded DER, since that’s what we need in the trusted root.

haydentherapper (approved):

PR #321, opened 2025-05-21
Fix gRPC and HTTP TLS connections on GCP

haydentherapper:
This mirrors how we've set up gRPC TLS in Fulcio, which is functional on GCP.

When gRPC TLS is configured, the HTTP connection does not need to be secure, since the connection is local. For gRPC, we will always return a constant certificate regardless of the handshake.

### DIFF
--- a/pkg/server/grpc.go
+++ b/pkg/server/grpc.go
@@ -141,8 +141,10 @@ func loadTLSCredentials(certFile, keyFile string) (credentials.TransportCredenti
 	}
 
 	tlsConfig := &tls.Config{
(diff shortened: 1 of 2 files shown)


haydentherapper:
Tested locally, HTTP is now working. gRPC is failing because macOS is very opinionated on TLS certificates, but given this works in GCP, I assume this is functional.

haydentherapper:
Actually I realized this might not work for gRPC, I think this has something to do with the healthcheck validating the TLS certificate. Going to try a different fix first.

PR #329, opened 2025-05-28
Change embedded version naming on proto types

loosebazooka:
...V0_0_2 -> ...V002

See naming: https://protobuf.dev/programming-guides/style/#identifier

We were initially using this to avoid collisions, but those are theoretical and we can work around them. Additionally we now can identify collisions at the proto spec level rather than at generation time.

#### Summary
Clients will once again need to regenerate type code. @jku we probably also want to disable pydantic here in a followup?


haydentherapper (approved):
Thanks!

PR #337, opened 2025-06-02
Merge proto defs into a single file

loosebazooka:
Should be easier to sync with protobuf-specs as a single file to add to the generator there.

Also prevents generation of unuseful openapi defs for files without any services.

unresolved: openapi still missing Entry (canonicalized_body). We could maybe create a NoOp service for that.


haydentherapper:
If we're doing this just for OpenAPI generation, that doesn't seem like a strong enough motivation. The messages across different proto files are logically divided. Clients shouldn't have to import the service definition to use request and response objects.

loosebazooka:
obselete

PR #347, opened 2025-06-09
Remove python client library builder

### LINKED ISSUE #338: Client generation and mirroring in protobuf-specs
For clients to make use of the rekor_v2 api, they would ideally have access to the protos or proto clients. To make this available, we would.
- mirror the protos on [`protobuf-specs`](https://github.com/sigstore/protobuf-specs) and generate clients
- keep the mirror up to date with the goings on of rekor-tiles.
- adjust the go-packages (or other language packages) to match protobuf specs outputs instead of the go layout we have in this project.

### Copy, Sync, Release Procedure:
1. Daily cron to sync change in `protobuf-specs` to what's in `rekor-tiles@v<latest>`
2. If changes detected, copy and make adjustments
    - go-package needs to change
    ```diff
    - option go_package = github.com/sigstore/rekor-tiles/pkg/generated/protobuf
    + option go_package = github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v2
    ```
    - other language package paths could probably be set at source of truth (rekor-tiles)
3. Rebuild protobuf-specs and prep for release.
    - Release a new patch version of protobuf specs vX.Y.(Z+1)
    - Or maybe a hyphenated release while we finalize the rekor-tiles api vX.Y.Z-tiles-A.B.C that matches the rekor-tiles version.

An alternative approach here would be to dynamically include rekor-tiles in the build and lock a reference to the rekor-tiles repository by hash. This approach seems a little more difficult to track/debug if something goes wrong.  

#### Option 1 (many files)
Currently set up as `<rekor-tiles>/api/protos/<many protos>` providing some nice logically separation of types in many files, a little more annoying to copy over as we need to edit the go_package in a few places. We could create a target directory in protobuf-specs for rekor v2 protos like `<protobuf-specs>/protos/rekor_v2/*.proto` or just dump all the protos in `./protos/.`.

##### Option 2 (merge into single file: like #337)
merge all current protos into `<rekor-tiles>/api/protos/rekor_v2.proto` providing a single file to sync and to represent the types specifically tied to rekor_v2. This may match the structure of `<protobuf-specs>/protos` a little better with a single file being imported for "rekor-v2"

haydentherapper:
Now moved to protobuf-specs.

Fixes #338.


haydentherapper:
Confirmed `make all` works as expected.

PR #377, opened 2025-06-13
Change default origin for test setup

cmurphy:
Clients need to deduce the checkpoint signer's origin in order to verify entries, and have no good way to accept a hint if the origin string does not match the server hostname the client is connecting to. This change updates the default hostname in the docker compose setup to match what most tests or local deployments will usually use to connect to Rekor.


ramonpetgrave64 (dismissed):

compose.yml
```diff
@@ -57,7 +57,7 @@ services:
     - "serve"
     - "--http-address=0.0.0.0"
     - "--grpc-address=0.0.0.0"
-    - "--hostname=rekor-local"
+    - "--hostname=localhost:3003"
```
ramonpetgrave64, line 60: The port number is part of the hostname? maybe rename this to "host"?

cmurphy:
> assuming the convention is a scheme-less URI?

Yes this is the problem, the clients reach the server by its URI, but here the hostname was not a URI and not related to the URI.

haydentherapper:
Sorry, not signing config, but the trusted root, which specifies a `baseURL`. We should update the documentation on https://github.com/sigstore/protobuf-specs/blob/7dafe972a1aefd1ab91127263e5d2c782a55977a/protos/sigstore_trustroot.proto#L37 for this convention.

In the trusted root, would you set `baseUrl` to `http://localhost:3003` or just `localhost`? I don't think we should set the precedent to drop the port from the URL as well. 

Just for context, where is this coming up?

haydentherapper:
What I'm suggesting is, verification shouldn't require online lookups, so `baseUrl` should now be unused. We could just use `baseUrl` as the origin string.

cmurphy:
Okay good idea :+1:

Issue #234, opened 2025-04-16
Move non-library packages under /internal/


haydentherapper:
**Description**

We need to decide what is a package that should be exported to callers as a library vs only needed within the service. The client and pkg/verify are needed by callers at least.

Additionally, after the move, document our public API

PR #390, opened 2025-06-24
Move internal packages to internal/

cmurphy:
Move all packages that don't need to be used by clients to an internal/ directory so that their APIs can be more freely updated.

Add godoc comments to the remaining exported names under pkg/ that had been missed before.

Fixes https://github.com/sigstore/rekor-tiles/issues/234


pkg/note/note.go
```diff
@@ -74,14 +74,17 @@ type noteVerifier struct {
...
 	return n.name
 }
 
+// Keyhash implements note.Verifier.
 func (n *noteVerifier) KeyHash() uint32 {
 	return n.hash
 }
 
+// Verify implements note.Verifier.
 func (n *noteVerifier) Verify(msg, sig []byte) bool {
```
ramonpetgrave64, line 88: We should re-use the original docs

 - https://pkg.go.dev/golang.org/x/mod/sumdb/note#Verifier

ramonpetgrave64 (changes requested):

internal/server/serve.go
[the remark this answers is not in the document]
haydentherapper: YAGNI - if a use case comes up, we can export it. There are no requests for this currently.

internal/verifier/verifier.go
haydentherapper, line 1: Just flagging that these may be exported later for monitoring, but fine to leave internal for now.

haydentherapper (approved):

ramonpetgrave64 (approved):

loosebazooka:
just curious, do I still need to track public doc comments? or does this cover everything?

cmurphy:
> or does this cover everything?

I think I got everything.

PR #392, opened 2025-06-26
Retract accidental tag

haydentherapper:
This adds a retract directive to prevent pkg.go.dev from serving the accidental v0.99.0 tag.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -172,3 +172,8 @@ tool (
 	github.com/google/addlicense
 	golang.org/x/tools/cmd/goimports
 )
+
+retract (
+	v0.99.1 // Contains only retractions
+	v0.99.0 // Accidental tag
+)


loosebazooka:
My understanding of the retract docs seems to indicate this will only work of we publish a version newer than 0.99.0

haydentherapper:
Thanks yall. Re-reading the docs, it looks like we could correct this if we cut v0.99.1 and retract both, which will then fallback to the last unretracted version. Any opposition to this?

haydentherapper:
Alternatively, we do nothing, and just ignore dependabot for cosign and Sigstore-go. Downside would be if any other Go clients are trying out the API.

haydentherapper:
I’m fighting plane internet, I’ll try this out in a bit. 

I think yes, based on https://go.dev/ref/mod#:~:text=When%20a%20user%20runs%20go%20get%20example.com/m@latest%2C%20the%20go%20command%20readsretractions%20from%20v1.0.1%2C%20which%20is%20now%20the%20highest%20version.%20Both%20v1.0.0%20andv1.0.1%20are%20retracted%2C%20so%20the%20go%20command%20will%20upgrade%20(or%20downgrade!)%20tothe%20next%20highest%20version%2C%20perhaps%20v0.9.5

loosebazooka:
Ah okay yeah I see it now:
```
As an example, consider a case where the author of module example.com/m publishes version v1.0.0 accidentally. To prevent users from upgrading to v1.0.0, the author can add two retract directives to go.mod, then tag v1.0.1 with the retractions.

retract (
    v1.0.0 // Published accidentally.
    v1.0.1 // Contains retractions only.
)

When a user runs go get example.com/m@latest, the go command reads retractions from v1.0.1, which is now the highest version. Both v1.0.0 and v1.0.1 are retracted, so the go command will upgrade (or downgrade!) to the next highest version, perhaps v0.9.5.


```

haydentherapper:
Updated to include v0.99.1 tag. After this is merged, I'll cut the tag and cancel the GHA workflow to prevent it from cutting a release.

Issue #158, opened 2025-03-28
Enable persistent antispam options in compose.yml and e2e tests


cmurphy:
**Description**

The "persistent" deduplication operator in Tessera does not work on the GCP Spanner emulator because it [does not implement](https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/172) the [batch write](https://cloud.google.com/spanner/docs/batch-write) feature. Enable it and test it when it is available. This does not block us from using it in real GCP.

cmurphy:
Blocked by https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/248

haydentherapper:
This might not be entirely fixed. I'm seeing the following error in the local log:

```
{
  "time": "2025-09-04T04:49:44.034978107Z",
  "severity": "ERROR",
  "logging.googleapis.com/sourceLocation": {
    "function": "github.com/transparency-dev/tessera/storage/gcp/antispam.(*follower).Follow",
    "file": "/opt/app-root/pkg/mod/github.com/transparency-dev/tessera@v1.0.0-rc2/storage/gcp/antispam/gcp.go",
    "line": 355
  },
  "message": "Failed to commit antispam population tx: failed to write antispam record: Transaction 1756922403123431 aborted due to active transaction 1756922403123430. The emulator only supports one transaction at a time. (Aborted)"
}
```

I'm running a local instance spun up from https://github.com/sigstore/scaffolding/tree/main/actions/setup-sigstore-env.

haydentherapper:
https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/266

PR #405, opened 2025-07-14
Enable persistent antispam option

cmurphy:
The missing feature in the Spanner emulator preventing Tessera from enabling deduplication functionality was resolved, so enable it by default in the docker compose file and add an end to end test for it.

Fixes https://github.com/sigstore/rekor-tiles/issues/158


haydentherapper (approved):
Thanks for following up on this!

PR #416, opened 2025-07-30
Update sharding for metrics and tinkey

haydentherapper:
Added a step to set up the monitoring (which can likely be applied earlier in the process, for now we'll do it after spinning up the service) and switching to using a script for generating Tink keysets

### DIFF
--- a/SHARDING.md
+++ b/SHARDING.md
@@ -331,7 +331,7 @@ You should have the log public key saved in `public.pem` from when you generated
 If you don't have the public key, look at the service logs to find the public key, which
 is logged on service startup.
 
-The key is PEM-encoded. Remove the PEM header and footer and remove all newline characters to get the DER-encoded public key.
+The key is PEM-encoded. Remove the PEM header and footer and remove all newline characters to get the base64-encoded PKIX public key.
 
 ### Update TrustedRoot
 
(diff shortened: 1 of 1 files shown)


SHARDING.md
```diff
@@ -307,13 +325,13 @@ have picked up the latest TrustedRoot.
...
 
-Look at the service logs to find the public key. This is the easiest way to retrieve
-the public key, as there is no CLI tool to convert a Tink keyset into a DER-encoded public key.
+You should have the log public key saved in `public.pem` from when you generated the key.
+If you don't have the public key, look at the service logs to find the public key, which
+is logged on service startup.
 
-The key will be printed as a PEM-encoded public key. Remove the PEM header and footer and remove
-all newline characters to get the DER-encoded public key.
+The key is PEM-encoded. Remove the PEM header and footer and remove all newline characters to get the DER-encoded public key.
```
jku, line 334: nit: I think calling it DER encoded is a bit misleading -- it's not binary like DER is supposed to be

jku (dismissed):
looks good to me. Left a comment on usage of "DER" that I guess I missed in the  original PR

SHARDING.md
haydentherapper: Thanks, updated to state this is base64-encoded PKIX, which is more precise.

Issue #370, opened 2025-06-11
Review pod configuration for healthiness


Hayden-IO:
**Description**

Review pod configuration for healthiness, including readiness and liveness probes and startup/teardown timeouts.

From https://victoriametrics.com/blog/go-graceful-shutdown/.

Hayden-IO:
@cmurphy I noticed that we didn't add the sleep binary to Dockerfile.release - was that intentional, or do we need to copy that in as well?

cmurphy:
I don't think it was intentional. Unfortunate side effect of having separate Dockerfiles.

Hayden-IO:
I'll update the release dockerfile now. I was thinking about just switching the release dockerfile to be the default and removing the other one, is anyone relying on delve?

cmurphy:
I've used it in the past and the dockerfile for the other services have it, which I think is why it was included here. But I can build my own image if I need it.

Hayden-IO:
https://github.com/sigstore/rekor-tiles/pull/926/ - I've dropped the release Dockerfile, but left a later stage for building with delve.

PR #432, opened 2025-08-08
Add better shutdown handling for tessera

cmurphy:
Before this change. it was possible for Tessera to get in the way of the process shutting down in two ways:

1) The context in the Appender did not respond to signals, so if it ran
   into an irrecoverable error in `integrateEntriesJob`, it would retry
   indefinitely and the process would not end shutdown.
2) The appender Shutdown function also runs in an indefinite loop, and
   if Tessera got into a corrupt state where it could never finish
   appending the sequenced entries, the shutdown loop would never
   terminate.

This change adds a new timeout parameter specifically for Tessera to give it a maximum time to finish processing entries. In the event it is stuck looping trying to process an unprocessable entry, it will now terminate in response to SIGINT/SIGTERM.

The default timeout for the server timeout, which is shared by the HTTP and gRPC servers, is reduced to 20s, which in a correctly functioning system with a checkpoint period of 10s should be long enough to finish processing all in-flight requests. The new tessera timeout is 30s. This means that in a badly behaving system where tessera did not integrate entries in time to send the response to clients, it still has 30s to try to finish committing entries to the log and empty the Spanner queue.

In the event that tessera encounters an irrecoverable error and cannot finish processing entries, the maximum time to shut down is 50s by default. This is reduced from 60s (server idle timeout) + indefinite (tessera loop). The Pod termination grace period should be extended to 50-60s accordingly.

Partial https://github.com/sigstore/rekor-tiles/issues/370

### DIFF
--- a/cmd/rekor-server/app/serve.go
+++ b/cmd/rekor-server/app/serve.go
@@ -154,7 +154,7 @@ var serveCmd = &cobra.Command{
 			server.NewHTTPConfig(
 				server.WithHTTPPort(viper.GetInt("http-port")),
 				server.WithHTTPHost(viper.GetString("http-address")),
-				server.WithHTTPTimeout(viper.GetDuration("timeout")),
+				server.WithHTTPTimeout(viper.GetDuration("server-timeout")),
 				server.WithHTTPMaxRequestBodySize(viper.GetInt("max-request-body-size")),
 				server.WithHTTPMetricsPort(viper.GetInt("http-metrics-port")),
 				server.WithHTTPTLSCredentials(viper.GetString("http-tls-cert-file"), viper.GetString("http-tls-key-file")),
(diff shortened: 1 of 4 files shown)


cmurphy:
> I assume this will solve some of the failures listed in #434?

No, I don't think so. This just makes it so that if the process needs to be killed it will actually be killed instead of getting stuck waiting forever.

cmd/rekor-server/app/serve.go
```diff
@@ -182,7 +183,8 @@ func init() {
 	serveCmd.Flags().Int("http-metrics-port", 2112, "HTTP port to bind metrics to")
 	serveCmd.Flags().Int("grpc-port", 3001, "GRPC port to bind to")
 	serveCmd.Flags().String("grpc-address", "127.0.0.1", "GRPC address to bind to")
-	serveCmd.Flags().Duration("timeout", 60*time.Second, "timeout")
+	serveCmd.Flags().Duration("server-timeout", 20*time.Second, "timeout settings for gRPC and HTTP connections")
+	serveCmd.Flags().Duration("log-timeout", 30*time.Second, "timeout for terminating the tiles log queue")
```
haydentherapper, line 187: nit: I might move this under `// tessera lifecycle configs` or rename this to mention Tessera, since there's a few other flags about logging

haydentherapper (dismissed):

PR #442, opened 2025-08-11
Catch pushback error

cmurphy:
Catch the error if `--pushback-max-outstanding` requests is reached so that the client can wait and retry.

Noticed while reviewing the code base as part of #370 that this error wasn't being handled as [shown in the tessera examples](https://github.com/transparency-dev/tessera/blob/72c95603719317e660ad50f29facc99c13263b3e/cmd/conformance/gcp/main.go#L103-L107).

I wasn't able to fully test this due to https://github.com/transparency-dev/tessera/issues/731.

### DIFF
--- a/internal/server/service.go
+++ b/internal/server/service.go
@@ -125,6 +125,9 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
 	}
 	entry := ttessera.NewEntry(canonicalized)
 	tle, err := s.storage.Add(ctx, entry)
+	if errors.Is(err, ttessera.ErrPushback) {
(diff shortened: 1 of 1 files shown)


internal/server/service.go
```diff
@@ -125,6 +125,9 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
 	}
 	entry := ttessera.NewEntry(canonicalized)
 	tle, err := s.storage.Add(ctx, entry)
+	if errors.Is(err, ttessera.ErrPushback) {
+		return nil, status.Errorf(codes.Unavailable, "reached max pushback; retry")
```
loosebazooka, line 129: should this be 429? instead of 503?

PR #461, opened 2025-08-21
Handle context cancel during entry upload

haydentherapper:
We've seen instances of client-side request cancellations from GitHub Actions, which has caused a bunch of unexpected 500s in the log.

When a context cancellation occurs, this will now be a client error (499) rather than an internal server error (500).

### DIFF
--- a/internal/server/service.go
+++ b/internal/server/service.go
@@ -128,6 +128,10 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
 	if errors.Is(err, ttessera.ErrPushback) {
 		return nil, status.Errorf(codes.Unavailable, "reached max pushback; retry")
 	}
+	if errors.Is(err, context.Canceled) {
+		// Returns a 499 Client Closed Request
+		return nil, status.Error(codes.Canceled, err.Error())
+	}
 	if errors.As(err, &tessera.DuplicateError{}) {
 		return nil, status.Error(codes.AlreadyExists, err.Error())
 	}
--- a/internal/server/service_test.go
+++ b/internal/server/service_test.go
@@ -21,14 +21,16 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/sigstore/protobuf-specs/gen/pb-go/dsse"
-
 	v1 "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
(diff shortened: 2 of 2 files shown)


haydentherapper:
FYI @mhutchinson

internal/server/service.go
```diff
@@ -128,6 +128,10 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
 	if errors.Is(err, ttessera.ErrPushback) {
 		return nil, status.Errorf(codes.Unavailable, "reached max pushback; retry")
 	}
+	if errors.Is(err, context.Canceled) {
+		// Returns a 499 Client Closed Request
+		return nil, status.Error(codes.Canceled, err.Error())
+	}
 	if errors.As(err, &tessera.DuplicateError{}) {
```
loosebazooka, line 135: just curious about something here, as we're mixing  `errors.Is` and `errors.As`. Why do we need to use `As` if we're not further using the typed converted error, could these all just be `Is`?

loosebazooka (approved):

internal/server/service.go
haydentherapper: Good catch, agreed we can use `As`. Updated!

internal/server/service.go
```diff
@@ -128,10 +128,14 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
 	if errors.Is(err, ttessera.ErrPushback) {
 		return nil, status.Errorf(codes.Unavailable, "reached max pushback; retry")
 	}
-	if errors.As(err, &tessera.DuplicateError{}) {
+	if errors.Is(err, context.Canceled) {
+		// Returns a 499 Client Closed Request
+		return nil, status.Error(codes.Canceled, err.Error())
+	}
+	if errors.Is(err, tessera.DuplicateError{}) {
```
loosebazooka, line 135: I think you can use the type directly, instead of the empty struct.

haydentherapper: Using just `tessera.DuplicateError` doesn't compile, though there's an issue with this change, the e2e tests are failing. Looking into it...

haydentherapper: I think `errors.Is` is for comparing an `error`, rather than `errors.As` which is for custom types that implement the `error` interface.

loosebazooka (approved):

Issue #56, opened 2025-02-24
Update API entrypoint summaries


haydentherapper:
**Description**

PR #500, opened 2025-09-09
Add metadata to Swagger output

cmurphy:
Use the protoc-gen-openapiv2 plugin to set some metadata on the Rekor Swagger JSON output so that the version no longer appears as "version not set" and the title is user-friendly.

Relates to https://github.com/sigstore/rekor-tiles/issues/56
Depends on https://github.com/sigstore/protobuf-specs/pull/757


haydentherapper (dismissed):
LGTM - @loosebazooka any comments?

Makefile
```diff
@@ -89,9 +89,10 @@ ko-local: ## Build container images locally using ko
 # generate Go protobuf code
 protos:
 	@echo "Generating go protobuf files"
+	go install tool github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2
 	@mkdir -p ${OPENAPI_OUT}
 	${DOCKER_RUN} -v ${PWD}:${MOUNT_POINT} ${SIGSTORE_PROTO_BUILDER} \
-		-I/opt/include -I/googleapis -I/protobuf-specs -I${MOUNT_POINT}/api/proto \
+		-I/opt/include -I/googleapis -I/protobuf-specs -I${MOUNT_POINT}/third_party/googleapis -I${MOUNT_POINT}/api/proto \
```
loosebazooka, line 95: I think these should probably come from the common docker image builder if we want to keep these up to date using the sigstore/protobuf-specs infra.

I don't think that has to block this, but we should file a bug and update protobuf-specs.

loosebazooka (approved):

haydentherapper:
Just needs to be rebased

Makefile
cmurphy: Can you give a more concrete example on where I should pull these from? I'd rather go ahead and get this right in this PR if I can.

loosebazooka (approved):

api/proto/rekor/v2/rekor_service.proto
```diff
@@ -29,6 +30,32 @@ option java_multiple_files = true;
...
+    };
+    license: {
+      name: "Apache License 2.0";
+      url: "https://github.com/sigstore/rekor-tiles/blob/main/LICENSE";
+    };
+  };
+
+  host: "*.rekor.sigstore.dev";
+  external_docs: {
+    url: "https://github.com/sigstore/rekor-tiles";
```
loosebazooka, line 50: Would this point to README.md? I dunno if that even matters?

loosebazooka (approved):

api/proto/rekor/v2/rekor_service.proto
cmurphy: The readme is the first thing you see when you visit that link

PR #526, opened 2025-09-25
Pipe `base64` to `tr -d '\n'` to clean up line breaks on linux

petertw9009:
#### Summary
Pass -w 0 to base64 to prevent it from adding line breaks since the v2 server rejects them.

-w 0 is not needed in the first base64 subshell command since the artifact_dgst is always short enough to fit in one line.

#### Release Note
Closes  sigstore/rekor-tiles/issues/525

### DIFF
--- a/CLIENTS.md
+++ b/CLIENTS.md
@@ -560,7 +560,7 @@ Post an entry:
 ```bash
 curl  -H \
  "Accept: application/json" -X \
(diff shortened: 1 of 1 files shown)


petertw9009:
Yeah i think you're right, the -w 0 is only required for linux. (and doesn't exist on macOS)

petertw9009:
One cross-platform suggestion from https://stackoverflow.com/questions/46463027/base64-doesnt-have-w-option-in-mac is to pipe the output into tr -d '\n'

loosebazooka:
It looks like we might be able to delete the '\n' one time at the end of the command?

Issue #523, opened 2025-09-22
Update package versions to v2


haydentherapper:
**Description**

@cmurphy suggested we cut a v2.0.0 release rather than v1.0.0, since this service and its libraries conceptually represent a v2 of Rekor, even if it's the v1 release of "rekor-tiles".

PR #544, opened 2025-10-06
Rename module to v2

cmurphy:
Fixes https://github.com/sigstore/rekor-tiles/issues/523


loosebazooka:
Is there a list somewhere of the imports we need to update outside this repo?

PR #558, opened 2025-10-16
Enable log witnessing

Hayden-IO:
The witness configuration takes in a policy as defined in https://git.glasklar.is/sigsum/core/sigsum-go/-/blob/main/doc/policy.md, which specifies the quorum of witnesses required for the log to produce a checkpoint.

For now, while witnesses are in development, we will fail open by default. This will be enabled only in the staging instance.


cmd/rekor-server/app/serve.go
```diff
@@ -148,6 +148,18 @@ var serveCmd = &cobra.Command{
 			}
 			appendOptions = tessera.WithLifecycleOptions(appendOptions, viper.GetUint("batch-max-size"), viper.GetDuration("batch-max-age"), viper.GetDuration("checkpoint-interval"), viper.GetUint("pushback-max-outstanding"))
 			appendOptions = tessera.WithAntispamOptions(appendOptions, persistentAntispam)
+			if wpf := viper.GetString("witness-policy-file"); wpf != "" {
+				f, err := os.ReadFile(wpf)
```
bobcallaway:
```suggestion
				f, err := os.ReadFile(filepath.Clean(wpf))
```

compose.yml
```diff
@@ -81,9 +82,39 @@ services:
...
     depends_on:
       rekor_init:
         condition: service_completed_successfully
+      witness:
+        condition: service_healthy
+
+  witness:
+    build:
+      # TODO: Change to transparency-dev/witness.git#main once PR is merged
+      context: https://github.com/haydentherapper/witness.git#additional-logs
```
loosebazooka: can you file a bug for this TODO?

loosebazooka:
curious what the affect on latency is once this witness stuff is added in. 

do we have conformance tests for witnessed checkpoints?

cmd/rekor-server/app/serve.go
```diff
@@ -235,6 +247,9 @@ func init() {
 	serveCmd.Flags().Uint("antispam-max-batch-size", 0, "maximum batch size for deduplication operations; will default to Tessera recommendation if unset; for Spanner, recommend around 1500 with 300 or more PU, or around 64 for smaller (e.g. 100 PU) instances")
 	serveCmd.Flags().Uint("antispam-pushback-threshold", 0, "maximum number of 'in-flight' add requests the antispam operator will allow before pushing back; will default to Tessera recommendation if unset")
 
+	// witness configs
+	serveCmd.Flags().String("witness-policy-file", "", "optional witness policy file")
```
cmurphy: nit - would be more consistent with other flags to call this `witness-policy-path`, or more concise to just call it `witness-policy`.

tests/testdata/witness/config.yaml
```diff
@@ -0,0 +1,5 @@
+Logs:
+  - Origin: rekor-local
+    URL: http://localhost:7080/tiles
```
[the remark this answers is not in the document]
haydentherapper: No, the witness isn't using the URL in this mode when `poll_interval = 0` - the URL would be used if the witness were polling for new checkpoints rather than receiving consistency proofs - but I'll use `gcs` still.

cmd/rekor-server/app/serve.go
haydentherapper: Looks like we have both `-path` and `-file`. I'll go with `path`.

internal/tessera/tessera.go
```diff
@@ -97,6 +97,17 @@ func WithAntispamOptions(opts *tessera.AppendOptions, as tessera.Antispam) *tess
 	return opts
 }
 
+func WithWitnessing(opts *tessera.AppendOptions, witnessPolicy []byte) (*tessera.AppendOptions, error) {
+	wg, err := tessera.NewWitnessGroupFromPolicy(witnessPolicy)
+	if err != nil {
+		return nil, fmt.Errorf("creating witness group from policy: %w", err)
+	}
+	// Don't block if witnesses are unavailable.
+	wOpts := &tessera.WitnessOptions{FailOpen: true}
```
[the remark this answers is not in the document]
haydentherapper: I thought about this. Given that there's no production witness network, I'd prefer to bake in that witnessing is always fail open for now. When it becomes more reliable, we can add a flag `--fail-open`, defaulting to `false` but still making it configurable for deployers that want to first test their configuration.

PR #573, opened 2025-10-30
Add support for AWS backend

kipz:
#### Summary

 This PR adds support for AWS as a storage backend for Rekor tiles

   **Key changes:**
   - Implemented AWS storage backend with:
     - AWS S3 for object storage (with support for S3-compatible services like MinIO)
     - Aurora MySQL / RDS MySQL** for database operations
   - Refactored e2e tests to use table-driven pattern for testing multiple backend configurations
   - Updated Makefile/build/CI to create different binaries for GCP and AWS 
     - `rekor-server-aws` - AWS-only dependencies
     - `rekor-server-gcp` - GCP-only dependencies

   **Testing:**

   For GCP backend:
   ```bash
   docker compose up -d --build --wait
   TEST_BACKENDS=gcp go test -v -tags=e2e ./tests/
   ```

   For AWS backend:
   ```bash
   docker compose -f docker-compose-aws.yml up -d --build --wait
   TEST_BACKENDS=aws go test -v -tags=e2e ./tests/
   ```

   For both backends:
   ```bash
   TEST_BACKENDS=all go test -v -tags=e2e ./tests/
   ```

   See [tests/README.md](tests/README.md) for more details.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,9 +31,9 @@ ADD ./internal/ $APP_ROOT/src/internal/
 
 ARG SERVER_LDFLAGS
 # Build server for deployment
-RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 go build -ldflags "${SERVER_LDFLAGS}" -o rekor-server ./cmd/rekor-server-${CLOUD_PROVIDER}
+RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 go build -tags ${CLOUD_PROVIDER} -ldflags "${SERVER_LDFLAGS}" -o rekor-server ./cmd/rekor-server-${CLOUD_PROVIDER}
 # Build server for debugger
-RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 go build -gcflags "all=-N -l" -ldflags "${SERVER_LDFLAGS}" -o rekor-server_debug ./cmd/rekor-server-${CLOUD_PROVIDER}
+RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 go build -tags ${CLOUD_PROVIDER} -gcflags "all=-N -l" -ldflags "${SERVER_LDFLAGS}" -o rekor-server_debug ./cmd/rekor-server-${CLOUD_PROVIDER}
 
 # Multi-stage deployment build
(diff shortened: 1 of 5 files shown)


tests/e2e_test.go
haydentherapper: Also will need to update where the test is invoked - https://github.com/sigstore/rekor-tiles/blob/eb8f925cee2c118b98d7dcbeca54123fc700908b/.github/workflows/test.yml#L115

haydentherapper: Can you also update the README? https://github.com/sigstore/rekor-tiles/blob/eb8f925cee2c118b98d7dcbeca54123fc700908b/tests/README.md

kipz:
> Drive-by comment: Have you considered having infra-specific binaries (e.g. `/cmd/rekor-server/gcp`, `/cmd/rekor-server/aws` etc.) as opposed to bundling support for everything into one super large binary?
> 
> For reference, this multi-binary approach is what TesseraCT does in order to avoid binary bloat from linking libs which are not actually used at runtime: https://github.com/transparency-dev/tesseract/tree/main/cmd/tesseract

I hadn't, but it seems like a good idea to me, thanks. What do others think? @haydentherapper ?

haydentherapper:
> > Drive-by comment: Have you considered having infra-specific binaries (e.g. `/cmd/rekor-server/gcp`, `/cmd/rekor-server/aws` etc.) as opposed to bundling support for everything into one super large binary?
> > For reference, this multi-binary approach is what TesseraCT does in order to avoid binary bloat from linking libs which are not actually used at runtime: https://github.com/transparency-dev/tesseract/tree/main/cmd/tesseract
> 
> I hadn't, but it seems like a good idea to me, thanks. What do others think? @haydentherapper ?

Thanks @AlCutter, that's a great suggestion. @kipz, can you take a look at doing this?

kipz:
> Just wanted to say that this is on my radar to review, this will just take a moment to do so! I'm happy with the overall direction though.

Thanks. I'll take look at those failures too :)

kipz:
> Thanks for getting this cleaned up! Since the PR has grown a bit large, I created a PR for refactoring the codebase to move the GCP bits and update the build scripts, #630. I took a slightly different approach to not use build tags and instead have the per-storage-backend dependencies under their own package. If this approach looks good to you, we can rebase this PR on top of 630. There's probably some more opportunity for deduplication, but for now some duplication between the GCP and AWS command impls seems fine.

Works for me. I don't like build tags (breaks my tools), and having a separate PR makes sense anyway!

haydentherapper:
https://github.com/sigstore/rekor-tiles/pull/643 for a checklist of the changes.

PR #574, opened 2025-11-01
verify: avoid huge tessera depenency for int64 cast

tonistiigi:
#### Summary

Remove dependency of tessera from the verify pkg. This single dependency import brings in 1923 additional files and increases binary size by 22MB, which seems excessive for integer type conversion.

For a full list of packages only brought in via this line compared to other imports already required by sigstore-go see https://gist.github.com/tonistiigi/aeaf17be2567074b07ed06ffd4713157 .

### DIFF
--- a/pkg/verify/verify.go
+++ b/pkg/verify/verify.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 
 	pbs "github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v1"
-	"github.com/sigstore/rekor-tiles/v2/internal/tessera"
 	f_log "github.com/transparency-dev/formats/log"
 	"github.com/transparency-dev/merkle/proof"
 	"github.com/transparency-dev/merkle/rfc6962"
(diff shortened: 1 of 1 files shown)


haydentherapper (commented):
Thanks for looking into this! Reducing the number of dependencies in sigstore-go has been a long-time goal, happy to see this significantly reduce the binary size.

Could we instead move safeint.go to its own package? It has no dependencies, so it should be fine to still use it, it just needs to be outside of the `tessera` package which pulls in the large GCP dependencies.

tonistiigi:
Maybe this could go into v0.1.11, with a different fix for master if that includes public API change?

For my taste, I don’t see why this `SafeInt` construction exists at all — or why anyone would ever create a log entry with a *negative* log index. An `int` is an `int`; it has one value. If some functions expect signed and others unsigned, the conversion should happen in between. There’s no such thing as a “safe int” existing in a quantum superposition of positivity and negativity.

Assuming this extra safety is needed on top of a regular cast at all, I’d suggest a small `safeint` package with:

```go
func ToUint64(int64) (uint64, error)
func ToInt64(uint64) (int64, error)
```

Note, though, that there are only two other calls to this package besides this line, so inline checks are probably fine. Or, if both remaining calls are in `internal/tessera`, a private helper would do.

If you *do* insist on keeping the `SafeInt` construction, I’d at least suggest making it:

```go
func NewSafeInt64[T int64 | uint64](number T) (*SafeInt64, error)
```

That would make it slightly more type-safe — as it stands, a simple typo when passing a variable could silently disable tlog verification at runtime.

Issue #452, opened 2025-08-14
Resource labels and custom metrics labels are missing from tessera metrics


cmurphy:
## Description

Tessera metrics are missing three of the six resource labels described here:

https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-otel#detect-resource-attributes

<img width="2174" height="506" alt="Image" src="https://github.com/user-attachments/assets/994c4f7f-9f0f-43e2-b23b-74cbb7cd219e" />

The [defined-by-us metric attribute `service.version`](https://github.com/sigstore/rekor-tiles/blob/fdc24378451350cadb26466e050b7e5b4b425105/cmd/rekor-server/app/otel.go#L42-L43) does not appear in these metrics. Though it looks like `service.name` is working, that might be getting populated by the [`PodMonitoring` config](https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-otel#:~:text=The%20Prometheus%20receiver%20automatically%20sets%20the%20service.name%20attribute%20based%20on%20the%20job_name%20in%20the%20scrape%20config%2C). Adding other attributes like [`Hostname`](https://pkg.go.dev/go.opentelemetry.io/otel@v1.37.0/semconv/v1.34.0#HostName) does not populate the label in the metrics explorer.

The attributes set at the [metric-level by tessera itself](https://github.com/transparency-dev/tessera/blob/d74ffe76145fa70eb4b71799bd81c0672abde55b/append_lifecycle.go#L419-L420) are working.

## Consequences

This means it is not possible to filter or aggregate tessera metrics by any of these attributes, which means it is impossible to tell when a metric arises from one shard or another or one revision or another:

<img width="1217" height="411" alt="Image" src="https://github.com/user-attachments/assets/a44388fa-a49c-4a71-a568-f3e885340344" />

There are also an extreme number of "TimeSeries could not be written...Points must be written in order. One or more of the points specified had an older start time than the most recent point." spamming the logs, which is another [potential effect](https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-otel#:~:text=Failure%20to%20set%20these%20labels%20to%20unique%20values%20can%20result%20in%20%22duplicate%20timeseries%22%20errors%20when%20exporting%20to%20Managed%20Service%20for%20Prometheus.) of this misconfiguration.

## Mitigation

[...]

PR #584, opened 2025-11-06
Switch OpenTelemetry exporter to Prometheus

cmurphy:
Make the metrics exportation consistent across metrics types. We already exported prometheus-style metrics to the /metrics endpoint, which the GCP Managed Prometheus collector understands. We don't need another GCP-specific exporter just for Tessera and OTel. Since GMP will now be reading these metrics, the resource labels will be applied properly in the metrics explorer view.

Partial or potentially complete fix for #452 - it won't be clear whether this resolves the out-of-order timeseries errors until we apply this and wait for a bit.

Thanks to @jku for correctly deducing the labeling problem and suggesting this solution.


jku (approved):
looks right to me

Issue #638, opened 2025-12-30
test coverage only counts unit tests (and nothing from tests/)


jku:
The coverage reporting is a bit useless since it only counts code in unit tests, and does not count the tests ins `tests/`.

I'll see if this is easy to fix -- at least e2e tests should be easy to include.

jku:
I've done a quick test and I think there is not enough value here: e.g. with e2e tests we'd like to get coverage on the service code but the actual e2e tests are client code and the server is running in docker ...

PR #639, opened 2025-12-30
workflows: Use codecov for e2e tests too

jku:
Let's see if this improves #638


jku:
I've not yet figured out why codecov does not like the files I produce.

Will see if it makes sense to continue with this: while looking at the tests I realized this approach is not _that_ useful: e.g. the gcp specific initialization in the "storage backend abstraction" PR won't be covered since it's not done in the test code (but in the container that the e2e test code uses)

PR #645, opened 2026-01-09
Add argument to make ko-local for handling different storage backends

Hayden-IO:
When creating another storage backend, I realized that we need to parameterize make ko-local rather than run multiple ko publish commands.

Also a little cleanup for the GCP backend: Updated the e2e test readme, unexported driver initialization functions (not a breaking change, they're under internal).


jku:
for some reason `TestServe_httpSmoke` timed out on the first run.

PR #661, opened 2026-01-21
Provide local proxy to mimic public instance

ret2libc:
Not sure if it was just me, but I had troubles using a local rekor-v2 for local development and I had to use nginx to "mimic" the public instance like `https://log2025-alpha2.rekor.sigstage.dev`.

Does it make sense to have this upstream?

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,40 @@ Run `docker compose up --build --wait` to start the service along with emulated 
 Run `docker compose down` to turn down the service, or `docker compose down --volumes` to turn down the service and delete
 persisted tiles.
 
+### Local deployment with nginx
+
+To deploy rekor-tiles locally with an nginx reverse proxy (serving on port 80):
+
+1. Add `rekor-local` to your hosts file:
+
+   ```bash
+   # Linux/macOS
+   echo "127.0.0.1 rekor-local" | sudo tee -a /etc/hosts
+
+   # Windows (run as Administrator)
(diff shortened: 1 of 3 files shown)


Hayden-IO (commented):
Is this meant as an alternative to using docker and serving the tiles through an emulated GCS bucket? Is that not sufficient for testing?

cmurphy:
It's currently in the regular config.yaml file for rekor-monitor https://github.com/sigstore/rekor-monitor/pull/854/changes#diff-f8656d2cdeaef2a8c20cd230a394a023d121771870063579ee0a8b5d11343b85L117-L131

I'm not opposed to standardizing it but it's so simple it seems a little overkill to me?

Hayden-IO:
The motivation to standardize it would be for other implementations of a log monitor. I'd propose an additional file with the list of logs, with each log including its read path, origin, and operator.

We could update the trusted root with an origin field as well, though we'd have to version it. I'd rather add a new config just for monitoring than maintain a second TR.

ret2libc:
> The motivation to standardize it would be for other implementations of a log monitor. I'd propose an additional file with the list of logs, with each log including its read path, origin, and operator.

What about the signing config url? Would you include that as well or was that excluded on purpose?

Hayden-IO:
With the new [monitoring config](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_monitor.proto), this should no longer be needed.

PR #744, opened 2026-04-03
Drop DSSE support

Hayden-IO:
We have found that DSSE requests can be quite large with comprehensive SBOMs, and these requests have become a source for server crashes in Rekor v1. Before Rekor v2 is widely deployed, we will drop support for DSSE requests, instead requiring clients to provide the DSSE PAE hash and signature extracted from the envelope.

We are not removing the protos to avoid any downstream breaking changes.


jku (dismissed):
looks correct to me

internal/server/metrics.go
```diff
@@ -68,12 +67,7 @@ var _initMetricsFunc = sync.OnceValue[*metrics](func() *metrics {
 
 	m.newHashedRekordEntries = f.NewCounter(prometheus.CounterOpts{
 		Name: "rekor_v2_new_hashedrekord_entries",
-		Help: "The total number of new dsse log entries",
```
loosebazooka, line 71: oh oops, that must look weird in our logs.

pkg/client/write/write.go
```diff
@@ -70,7 +70,7 @@ func NewWriter(writeURL string, opts ...client.Option) (Client, error) {
 	}, nil
 }
 
-// Add uploads a hashedrekord or DSSE log entry and returns the TransparencyLogEntry proving the entry's inclusion in the log.
+// Add uploads a hashedrekord log entry and returns the TransparencyLogEntry proving the entry's inclusion in the log.
```
loosebazooka, line 80: I'm a little curious how this is called. Should we leave this func signature generic to allow for some future flexibility, or should we try to change
```diff
- func (w *writeClient) Add(ctx context.Context, entry any) (*pbs.TransparencyLogEntry, error) {
+ func (w *writeClient) Add(ctx context.Context, entry pb.HashedRekorRequestV002) (*pbs.TransparencyLogEntry, error) {
```

loosebazooka (commented):
looks fine to me

loosebazooka (dismissed):

loosebazooka:
remember to resync the protobuf-specs import of these proto defs

pkg/client/write/write.go
cmurphy: Probably leave it generic, we're likely to add new ones

internal/server/service.go
```diff
@@ -104,24 +103,11 @@ func (s *Server) CreateEntry(ctx context.Context, req *pb.CreateEntryRequest) (*
...
-			Kind:    entry.Kind,
-			Version: entry.ApiVersion,
-		}
-		serialized, err = protojson.Marshal(entry)
-		if err != nil {
-			slog.WarnContext(ctx, "failed marshaling dsse request", "error", err.Error())
-			return nil, status.Errorf(codes.InvalidArgument, "invalid dsse request")
-		}
-		metricsCounter = getMetrics().newDsseEntries
+		_ = grpc.SetHeader(ctx, metadata.Pairs(httpStatusCodeHeader, "405"))
```
cmurphy: Is 405 the best code for this? The method and the endpoint are valid, it's just an unsupported type. What about reusing `codes.InvalidArgument` for this, as below?

Hayden-IO: Updated to use invalid argument

pkg/client/write/write.go
Hayden-IO: agreed on leaving it generic.

api/proto/rekor/v2/dsse.proto
```diff
@@ -29,17 +29,22 @@ option java_outer_classname = "RekorV2Dsse";
 option ruby_package = "Sigstore::Rekor::V2";
 
 // A request to add a DSSE v0.0.2 entry to the log
```
loosebazooka, line 31: should we remove this message?

CLIENTS.md
```diff
@@ -130,46 +84,17 @@ The client MUST gracefully fail when the client is not aware of a `kind` and `ap
...
-
-### Two Entry Types
+### Entry Type
 
-Rekor v2 only supports `hashedrekord` (`HashedRekordLogEntryV002`) and
-`dsse` (`DSSELogEntryV002`) entry types, dropping a number of unused types
-such as `jar`, `alpine`, `rpm`, and the older types `rekord` and `intoto`.
+Rekor v2 only supports `hashedrekord` (`HashedRekordLogEntryV002`), dropping
+a number of unused types such as `jar`, `alpine`, `rpm`, and the older
+types `rekord` and `intoto`.
```
loosebazooka:
```suggestion
types `rekord`, `intoto` and `dsse`.
```

loosebazooka (dismissed):
Is there any point in this middle state we're in? Should we just take the dsse stuff out completely?

api/proto/rekor/v2/dsse.proto
Hayden-IO: I'm going to leave this to avoid this being an API-breaking change for sigstore-go, which did use this type. In practice, I doubt it'll affect anyone, but let's be safe. Will drop in a future revision.

loosebazooka (approved):

PR #767, opened 2026-04-21
Bump go-tuf version

Hayden-IO:
This allows us to downgrade back down to 1.25.0 as the minimum Go version. We'll follow up with downgrading sigstore-go, which depends on this package.


Hayden-IO:
Note I have to bump the indirect dep because sigstore-go depends on this package (in various packages), and this library depends on sigstore-go (for tests).

loosebazooka (approved):

PR #787, opened 2026-05-19
Add ToEntryHash and VerifyLogEntryWithHash

codysoyland:
## Summary

Adds two functions that let a verifier reconstruct a hashedrekord entry hash from bundle-signed inputs, rather than trusting the persisted `canonicalized_body`:

- `pkg/types/hashedrekord.ToEntryHash(digest, *pb.Signature) ([]byte, error)` — reconstructs a `HashedRekordLogEntryV002` and runs the same `protojson.Marshal` → `jsoncanonicalizer.Transform` → `rfc6962.HashLeaf` pipeline used server-side. `HashOutput.Algorithm` is derived from the verifier's `KeyDetails` via the algorithm registry.
- `pkg/verify.VerifyLogEntryWithHash(entry, verifier, entryHash)` — verifies the checkpoint signature and inclusion proof against a caller-supplied entry hash instead of hashing `entry.CanonicalizedBody`.

Spec PR: sigstore/architecture-docs#63


pkg/verify/verify.go
```diff
@@ -68,6 +68,25 @@ func VerifyLogEntry(entry *pbs.TransparencyLogEntry, verifier sumdb_note.Verifie
 	return VerifyInclusionProof(entry, cp)
 }
 
+// VerifyLogEntryWithLeafHash is like VerifyLogEntry but uses a caller-provided leaf hash
```
Hayden-IO: Can we write this comment to be independent of other functions, since we might deprecate the other one? Let's also not reference the spec, that's prone to get out of sync.

pkg/verify/verify_test.go
```diff
@@ -307,6 +316,169 @@ func TestVerifyLogEntry(t *testing.T) {
...
+
+// TestRecomputeLeafEndToEnd drives the rekor-v2-spec §6.1.4 "Recompute the
+// leaf" path producer-to-consumer: a HashedRekordRequestV002 is canonicalized
+// the same way the log service does, the resulting leaf hash anchors a signed
+// size-1 checkpoint, and the bundle-signed inputs are then re-fed through
+// hashedrekord.ReconstructLeafHash + VerifyLogEntryWithLeafHash. This catches
+// any future divergence between the client reconstruction and server
+// canonicalization (protojson field ordering, defaults, jsoncanonicalizer
+// behavior, etc.) that the isolated unit tests would silently miss.
+func TestRecomputeLeafEndToEnd(t *testing.T) {
```
Hayden-IO: I don't think this test should be here, this is more than a unit test of a function.

pkg/types/hashedrekord/leafhash.go
```diff
@@ -0,0 +1,66 @@
...
+// The result is the leaf hash a Verifier can use as the input to an inclusion
+// proof, implementing the spec's preferred "Recompute the leaf" path
+// (rekor-v2-spec §6.1.4).
+//
+// HashOutput.Algorithm is derived from signature.Verifier.KeyDetails via the
+// Algorithm Registry, matching the server's canonicalization. Callers MUST
+// supply the bundle's digest and *pb.Signature directly; the verifier oneof
+// (X509Certificate vs PublicKey) inside signature.Verifier must match the
+// bundle's verification material.
+func ReconstructLeafHash(digest []byte, signature *pb.Signature) ([]byte, error) {
```
Hayden-IO: Would prefer a name aligned with other functions, like `ToEntryHash`? Let's not mention leaves, that's an impl detail.

pkg/types/hashedrekord/leafhash.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
+	v1 "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+	pb "github.com/sigstore/rekor-tiles/v2/pkg/generated/protobuf"
+	sigsignature "github.com/sigstore/sigstore/pkg/signature"
+	"github.com/transparency-dev/merkle/rfc6962"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// ReconstructLeafHash builds a HashedRekordLogEntryV002 from the supplied
```
Hayden-IO: Can we make this comment less verbose? This is explaining the implementation, not the why of the function.

pkg/types/hashedrekord/leafhash.go
Hayden-IO, line 1: Can this be in the existing hashedrekord.go file?

pkg/verify/verify.go
```diff
@@ -68,6 +68,25 @@ func VerifyLogEntry(entry *pbs.TransparencyLogEntry, verifier sumdb_note.Verifie
...
+// instead of hashing entry.CanonicalizedBody. Use this when the caller reconstructs the
+// canonicalized entry from independently-verified inputs (see rekor-v2-spec §6.1.4
+// "Recompute the leaf").
+func VerifyLogEntryWithLeafHash(entry *pbs.TransparencyLogEntry, verifier sumdb_note.Verifier, leafHash []byte) error { //nolint: revive
```
Hayden-IO: Same as above, let's refer to this as an EntryHash rather than Leaf.

Hayden-IO (changes requested):

pkg/verify/verify.go
codysoyland: I updated it to `VerifyLogEntryWithHash`. I can change it to `VerifyLogEntryWithEntryHash` if you prefer but I felt that `Entry` in the name twice was stuttery.

Hayden-IO (approved):
Thanks!

kommendorkapten (approved):
Looks good!

PR #789, opened 2026-05-20
Set IntegratedTime in TransparencyLogEntry from storage.Add

### LINKED ISSUE #788: TransparencyLogEntry.IntegratedTime is never populated by tessera.Add
**Description**

`TransparencyLogEntry.IntegratedTime` is never populated in the response returned by `internal/tessera/tessera.go::storage.Add`. Verified on v2.2.1 and current `main` (HEAD as of 2026-05-20).

The relevant code:

https://github.com/sigstore/rekor-tiles/blob/main/internal/tessera/tessera.go#L147-L155

```go
return &rekor_pb.TransparencyLogEntry{
    LogIndex:          idx.I(),
    InclusionProof:    inclusionProof,
    CanonicalizedBody: entry.Data(),
}, nil
```

`IntegratedTime` is left at the proto default (`0`), and is serialized as `"integratedTime": "0"` in the HTTP JSON response.

This is a known gap — issue [#285](https://github.com/sigstore/rekor-tiles/issues/285) tracked all three missing fields (`LogID`, `KindVersion`, `IntegratedTime`). PR #292 (commit 6ab405) closed it by adding `LogID` and `KindVersion` is set on the `Server.CreateEntry` post-`storage.Add` path, but `IntegratedTime` was never landed and the issue was closed prematurely.

**Reproduction**

Submit any DSSE entry via `POST /api/v2/log/entries` against a fresh rekor-tiles v2.2.1 instance (Tessera POSIX backend) and inspect the response:

```json
{
  "logIndex": "1",
  "logId": { "keyId": "<base64>" },
  "kindVersion": { "kind": "dsse", "version": "0.0.2" },
  "integratedTime": "0",            ← expected something like "1779299049"
  "inclusionProof": { ... },
  "canonicalizedBody": "<base64>"
}
```

**Why it matters**

Sigstore bundle v0.3 consumers (cosign, sigstore-python, sigstore-js, custom verifiers) read `verificationMaterial.tlogEntries[0].integratedTime` to validate replay windows and timestamp witnessing. A value of `0` either fails strict verification or silently bypasses time-bound checks.

**Suggested fix**

Set `IntegratedTime` to the current Unix time at the point of integration in `storage.Add`, since the function only returns after the entry is integrated (the Tessera awaiter blocks until durable). This matches the semantic of "the time the log entry was integrated" per the protobuf spec.

```go
return &rekor_pb.TransparencyLogEntry{
    LogIndex:          idx.I(),
    InclusionProof:    inclusionProof,
    CanonicalizedBody: entry.Data(),
    IntegratedTime:    time.Now().Unix(),
}, nil
```

For higher fidelity, the timestamp could come from Tessera's integration tracker if exposed.

**Version**

rekor-tiles v2.2.1 (also reproduced against current `main`).

**Workaround**

Downstream consumers (we hit this in a private wiki-audit pipeline using rekor-tiles as a self-hosted Sigstore Rekor for DSSE) stamp `integratedTime` client-side at submit-response receipt since the response is the integration ack. This is a 99.9%-accurate proxy but technically inaccurate — the canonical fix belongs server-side.

3-Deacon:
Closes #788. Follows up on #285 — finishes the IntegratedTime half that PR #292 left behind.

## Summary

`internal/tessera/tessera.go::storage.Add` left `TransparencyLogEntry.IntegratedTime` at the proto default (0), serializing to `"integratedTime": "0"` in HTTP/JSON responses. Sigstore Bundle v0.3 consumers (cosign, sigstore-python, sigstore-js, custom verifiers) read this field to validate replay windows and timestamp witnessing; a 0 value either fails strict verification or silently bypasses time-bound checks.

`storage.Add` is synchronous on Tessera integration (`s.addEntry` awaits the `PublicationAwaiter`; `buildProof` requires the integrated tile state), so `time.Now().Unix()` at this point matches the protobuf-specs definition of "the time the log entry was integrated into the log".

## Background

Issue #285 tracked all three missing TransparencyLogEntry fields (`LogID`, `KindVersion`, `IntegratedTime`). PR #292 closed it by landing `LogID` (in `Server.CreateEntry`) and `KindVersion`, but `IntegratedTime` was never landed and the issue was closed prematurely. I filed #788 from the consumer side after observing `"integratedTime": "0"` in production bundles produced by v2.2.1.

## Change

```go
return &rekor_pb.TransparencyLogEntry{
    LogIndex:          idx.I(),
    InclusionProof:    inclusionProof,
    CanonicalizedBody: entry.Data(),
+   IntegratedTime:    time.Now().Unix(),
}, nil
```

`time` is already imported.

## Test plan

- [x] New `TestAddSetsIntegratedTime` bounds the returned value between `time.Now()` before and after `storage.Add` returns; verifies `IntegratedTime != 0`.
- [x] Existing `TestAdd` (success / integration_failed / duplicate_entry) still passes.
- [x] Full `go test ./...` passes (`internal/server`, `pkg/verify`, `pkg/client/write`, `pkg/note`, etc. all green).

## Release notes

Bundle consumers that previously had to compensate for `integrated_time = 0` (either by stamping client-side at submission or by treating the field as untrusted) can now rely on the server value.

### DIFF
--- a/internal/tessera/tessera.go
+++ b/internal/tessera/tessera.go
@@ -148,6 +148,16 @@ func (s *storage) Add(ctx context.Context, entry *tessera.Entry) (*rekor_pb.Tran
 		LogIndex:          idx.I(),
 		InclusionProof:    inclusionProof,
 		CanonicalizedBody: entry.Data(),
+		// Set IntegratedTime at the point the entry was sequenced and the
+		// inclusion proof was built. This function is synchronous on Tessera
+		// integration (s.addEntry awaits the awaiter; buildProof requires the
+		// integrated tile state), so time.Now() is the integration time. Per
+		// the protobuf-specs comment on TransparencyLogEntry.integrated_time,
+		// this field is the "time the entry was integrated into the log".
+		// Closes #285 follow-up (the integrated_time half of the
+		// "Complete returned TransparencyLogEntry" trio that PR #292 only
+		// partially landed).
+		IntegratedTime: time.Now().Unix(),
 	}, nil
 }
 
--- a/internal/tessera/tessera_test.go
+++ b/internal/tessera/tessera_test.go
@@ -96,6 +96,40 @@ gb/AnEEsBNpTobDduU3OSNaiTp6liYf31FoE6AB/s8o=
 	}
 }
 
+// TestAddSetsIntegratedTime guards against the regression where
+// TransparencyLogEntry.IntegratedTime is left at the proto default (0) — the
+// gap behind issue #285 that PR #292 only partially closed. Verifies the
+// returned value is a plausible recent Unix timestamp.
+func TestAddSetsIntegratedTime(t *testing.T) {
+	ctx := context.Background()
+	tileHash := hexDecodeOrDie(t, "81bfc09c412c04da53a1b0ddb94dce48d6a24e9ea58987f7d45a04e8007fb3ca")
+	readCheckpoint := func(_ context.Context) ([]byte, error) {
+		<-time.After(5 * time.Millisecond)
+		return []byte(`test.origin
+1
+gb/AnEEsBNpTobDduU3OSNaiTp6liYf31FoE6AB/s8o=
+
+— test.origin AAAAAW5vb3AKMQpnYi9BbkVFc0JOcFRvYkRkdVUzT1NOYWlUcDZsaVlmMzFGb0U2QUIvczhvPQo=`), nil
+	}
+	s := storage{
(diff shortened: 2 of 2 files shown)


Hayden-IO:
This is intentional and working as intended, the log is not a trusted source of time, and unlike the Trillian-backed Rekor instance where the timestamp was signed and bound to the logged entry.

We'd like to remove integrated time from the bundle in a future revision, but weren't interested in cutting another bundle version at this time, so for now this will remain unset.

Issue #827, opened 2026-06-10
Support ML-DSA in Rekor v2


Hayden-IO:
**Description**

As outlined in the [Sigstore PQC plan](https://docs.google.com/document/d/1g3WQk4JSwQ26MGwB-Kkhu8ylrPF5JnE4v62d2tDC2xY/edit?tab=t.0), Rekor will support ML-DSA. To support the now large signatures and public keys, Rekor will add a new entry type that stores hashes rather than pre-images. This concurrently will mitigate the risk of log poisoning.

This issue tracks support.

Tasks:
- [ ] Implement https://github.com/C2SP/C2SP/pull/244 as a new entry type (resolving https://github.com/sigstore/rekor-tiles/issues/75)
- [ ] Add a distinct server implementation that only supports uploading the identity entry type
    - [ ] Start with the POSIX storage backend
    - [ ] Response must be a c2sp.org/tlog-proof string
    - [ ] Implement a request API that supports ML-DSA signatures as a credential
    - [ ] Implement a request API that supports OIDC tokens as a credential 
    - [ ] Implement support for ML-DSA OIDC tokens (likely won't happen til Go 1.27)
    - [ ] Support ML-DSA signed log checkpoints (built into Tessera)
- [ ] Add end-to-end testing
- [ ] Add a GCP server implementation. We won't add AWS or GCP CloudSQL until asked.
- [ ] Implement out of tree receipt storage
   - [ ] Store signatures
   - [ ] Store some part of the identity token. Can't store the ID token signature, as that is effectively a password, but could store the unredacted provenance in a private bucket.

PR #828, opened 2026-06-10
Implement Identity-based Transparency Logging

Hayden-IO:
This change begins the implementation for the Identity-based Transparency Logging Server.

This change supports the new entry type, with only Ed25519 signatures as credentials. This will be proceeded by support for OIDC tokens as credentials, and ML-DSA signatures as credentials.

Partially implements https://github.com/sigstore/rekor-tiles/issues/827, with the remaining PRs to support ML-DSA and an additional GCP backend for PGI deployment.

Coding assisted with Gemini, reviewed entirely by me.

### DIFF
--- a/api/proto/rekor/v2/identity.proto
+++ b/api/proto/rekor/v2/identity.proto
@@ -32,6 +32,10 @@ message PublicKeyCredential {
     // The signature computed over the specification identifier, checksum, and context strings,
     // as defined by the c2sp.org/identity-transparency specification.
     bytes signature = 2 [(google.api.field_behavior) = REQUIRED];
+
+    // Optional context value to record alongside the message. Must be a SHA-256 digest.
+    // Must be signed as defined by the c2sp.org/identity-transparency specification.
+    bytes context = 3;
 }
 
 // A request to add an identity-based transparency log entry (v0.0.1)
@@ -42,7 +46,4 @@ message IdentityRequestV001 {
     
     // Must be a SHA-256 digest of data.
     bytes message = 3 [(google.api.field_behavior) = REQUIRED];
-    
-    // Optional context value. Must be a SHA-256 digest.
-    bytes context = 4;
 }
--- a/identity-posix-compose.yml
+++ b/identity-posix-compose.yml
@@ -1,0 +1,94 @@
+# Copyright 2026 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+services:
+  rekor:
+    build:
+      context: .
+      target: deploy
+      args:
+        STORAGE_BACKEND: posix
+    command:
+    - "rekor-server"
+    - "serve-identity"
+    - "--http-address=0.0.0.0"
+    - "--grpc-address=0.0.0.0"
+    - "--hostname=rekor-local"
+    - "--storage-dir=/tmp/identityposixlog"
+    - "--signer-filepath=/pki/ed25519-priv-key.pem"
+    - "--checkpoint-interval=2s"
+    - "--log-level=debug"
+    - "--request-response-logging=true"
+    - "--persistent-antispam=true"
+    - "--witness-policy-path=/witness/policy.yaml"
+    ports:
+    - "3006:3000" # http port
+    - "3007:3001" # grpc port
+    - "2115:2112" # metrics port
+    healthcheck:
+      test:
+      - CMD-SHELL
+      - curl http://localhost:3000/healthz | grep '{"status":"SERVING"}'
+      timeout: 30s
+      retries: 10
+      interval: 3s
+    volumes:
+    - ./tests/testdata/pki:/pki
+    - ./tests/testdata/witness:/witness
+    - identity_posix_storage:/tmp/identityposixlog
+    depends_on:
+      nginx:
+        condition: service_healthy
+      witness:
+        condition: service_healthy
+  nginx:
+    image: nginx:1.31.1@sha256:5aca99593157f4ae539a5dec1092a0ad8762f8e2eb1789085a13a0f5622369f6
+    volumes:
+      - identity_posix_storage:/usr/share/nginx/html
+    restart: always
+    ports:
+      - "8001:80"
+    healthcheck:
+      test: ["CMD", "curl", "--fail", "http://localhost/"]
+      interval: 30s
+      retries: 10
+      timeout: 3s
+  witness:
+    build:
+      context: https://github.com/transparency-dev/witness.git#main
+      dockerfile: cmd/omniwitness/Dockerfile
+    volumes:
+      - witness_data:/witness_data:rw
+      - ./tests/testdata/witness:/witness_config
+    command:
+      - "--listen=:8100"
+      - "--db_file=/witness_data/witness.sqlite"
+      - "--private_key_path=/witness_config/private.key"
+      - "--additional_logs=/witness_config/config.yaml"
+      - "--logtostderr"
+      - "--v=2"
+      - "--rate_limit=100"
+    restart: always
+    healthcheck:
+      test:
+      - CMD-SHELL
+      - wget --spider --tries=1 http://localhost:8081/metrics || exit 1
+      timeout: 30s
+      retries: 10
+      interval: 3s
+    ports:
+      - "8101:8100"
+volumes:
+  identity_posix_storage: {}
+  witness_data: {}
--- a/internal/server/grpc.go
+++ b/internal/server/grpc.go
@@ -39,7 +39,6 @@ import (
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
 	"github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/recovery"
-	pb "github.com/sigstore/rekor-tiles/v2/pkg/generated/protobuf"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
@@ -52,11 +51,11 @@ import (
 type grpcServer struct {
 	*grpc.Server
 	serverEndpoint string
-	serverImpl     any
+	serverImpl     HTTPRegistrar
 }
 
 // newGRPCServer starts a new grpc server and registers the services.
-func newGRPCServer(config *GRPCConfig, server any) *grpcServer {
+func newGRPCServer(config *GRPCConfig, server Registrar) *grpcServer {
 	var opts []grpc.ServerOption
 
 	grpcPanicRecoveryHandler := func(p any) (err error) {
@@ -88,15 +87,8 @@ func newGRPCServer(config *GRPCConfig, server any) *grpcServer {
 
 	s := grpc.NewServer(opts...)
 
-	if r, ok := server.(pb.RekorServer); ok {
-		pb.RegisterRekorServer(s, r)
-	}
-	if i, ok := server.(pb.IdentityRekorServer); ok {
-		pb.RegisterIdentityRekorServer(s, i)
-	}
-	if h, ok := server.(grpc_health_v1.HealthServer); ok {
-		grpc_health_v1.RegisterHealthServer(s, h)
-	}
+	server.RegisterGRPC(s)
+	grpc_health_v1.RegisterHealthServer(s, server)
 
 	reflection.Register(s)
 
--- a/internal/server/identity_service.go
+++ b/internal/server/identity_service.go
@@ -16,13 +16,25 @@ package server
 
 import (
 	"context"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"log/slog"
+	"strconv"
 
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"github.com/sigstore/rekor-tiles/v2/internal/safeint"
 	"github.com/sigstore/rekor-tiles/v2/internal/tessera"
 	pb "github.com/sigstore/rekor-tiles/v2/pkg/generated/protobuf"
+	"github.com/sigstore/rekor-tiles/v2/pkg/types/identity"
 	"github.com/sigstore/sigstore/pkg/signature"
+	"github.com/transparency-dev/formats/proof"
+	ttessera "github.com/transparency-dev/tessera"
 	"google.golang.org/genproto/googleapis/api/httpbody"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 )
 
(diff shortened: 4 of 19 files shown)


api/proto/rekor/v2/rekor_service.proto
Hayden-IO, line 61: Look at second service rather than second RPC

internal/server/grpc.go
```diff
@@ -86,8 +87,17 @@ func newGRPCServer(config *GRPCConfig, server rekorServer) *grpcServer {
 	}
 
 	s := grpc.NewServer(opts...)
-	pb.RegisterRekorServer(s, server)
-	grpc_health_v1.RegisterHealthServer(s, server)
+
+	if r, ok := server.(pb.RekorServer); ok {
```
Hayden-IO: @loosebazooka Do you think this looks clearer? There's no longer a common interface because the `CreateEntry` RPC differs between each service. This means having to dynamically determine types.

```
type rekorServer interface {
	pb.RekorServer
	grpc_health_v1.HealthServer
}
```

is now removed.

I'm going through old PRs, I see https://github.com/sigstore/rekor-tiles/pull/84/ but I can't find any place where we built an implementation with two services.

internal/server/serve.go
```diff
@@ -35,6 +38,18 @@ func Serve(ctx context.Context, hc *HTTPConfig, gc *GRPCConfig, tesseraTimeout t
 		os.Exit(1)
 	}
 
+	_, isRekor := s.(pb.RekorServer)
```
Hayden-IO: @loosebazooka This is what I've settled on. I looked at duplicating the server files but it created so much duplication that it seemed cleaner to just check the type assertions before initializing the server. This is all internal code and we're not going to add another server anyways.

api/proto/rekor/v2/identity.proto
```diff
@@ -0,0 +1,56 @@
...
+    bytes signature = 2 [(google.api.field_behavior) = REQUIRED];
+}
+
+// A request to add an identity-based transparency log entry (v0.0.1)
+message IdentityRequestV001 {
+    oneof credential {
+        PublicKeyCredential public_key = 1;
+    }
+    
+    // Must be a SHA-256 digest of data.
```
loosebazooka, line 47: just so I'm clear here, this not subject to the cryptographic agility mandate with respect to hash algorithm selection?

internal/server/grpc.go
```diff
@@ -86,8 +87,17 @@ func newGRPCServer(config *GRPCConfig, server rekorServer) *grpcServer {
...
+		pb.RegisterRekorServer(s, r)
+	}
+	if i, ok := server.(pb.IdentityRekorServer); ok {
+		pb.RegisterIdentityRekorServer(s, i)
+	}
+	if h, ok := server.(grpc_health_v1.HealthServer); ok {
+		grpc_health_v1.RegisterHealthServer(s, h)
+	}
+
```
loosebazooka: Type checking occurs before this call, but it appears we're not covering the full scope of the `any` on the function signature. Does it make sense to have some redundant runtime check here (or moving the other check from serve.go here) where the `any` is actually consumed.

internal/server/identity_service.go
```diff
@@ -0,0 +1,145 @@
...
+		ExtraData:  extraData,
+	}
+	proofBytes := p.Marshal()
+
+	return &httpbody.HttpBody{
+		ContentType: "text/plain",
+		Data:        proofBytes,
+	}, nil
+}
+
```
loosebazooka, line 131: in the posix case I thought we might be requried to implement these functions?

internal/server/identity_service.go
```diff
@@ -0,0 +1,145 @@
...
+	var extraData []byte
+	if len(req.GetContext()) > 0 {
+		var contextPairs []string
+		for _, entry := range req.GetContext() {
+			k := hex.EncodeToString(entry.GetKey())
+			v := hex.EncodeToString(entry.GetValue())
+			contextPairs = append(contextPairs, fmt.Sprintf("%s:%s", k, v))
+		}
+		extraData = []byte(strings.Join(contextPairs, "\n"))
```
loosebazooka: should this be provided to the user in sorted order?

loosebazooka (commented):
Could this maybe be split up into
1. empty server stub
2. createEntry implementation?

internal/server/identity_service.go
loosebazooka: oh never mind, we wont have any monitors or anything yet.

internal/server/identity_service.go
Hayden-IO: We can for consistency, although I don't think we should treat this like the canonicalized body - a client shouldn't assume they're sorted.

api/proto/rekor/v2/identity.proto
Hayden-IO: Correct. Given that the log specification hardcodes SHA-256 for the node hashes, there's no value in allowing flexibility with the request data digest algorithm. If a flaw is found in SHA-256, the log spec would have to be updated, at which point we'd update the leaf spec as well.

internal/server/identity_service.go
Hayden-IO: Should we just delete these RPCs? Even for POSIX, the log shouldn't implement them, they should be served out of band by an HTTP web server.

They were added originally when we thought we might support MySQL for both the write and read paths, there's no plans to do so anymore now that there's a Tessera backend for MySQL + bucket.

cmd/rekor-server/identity-posix/app/serve.go
```diff
@@ -0,0 +1,195 @@
...
+
+	"github.com/sigstore/rekor-tiles/v2/internal/algorithmregistry"
+	"github.com/sigstore/rekor-tiles/v2/internal/cli"
+	"github.com/sigstore/rekor-tiles/v2/internal/server"
+	"github.com/sigstore/rekor-tiles/v2/internal/signerverifier"
+	"github.com/sigstore/rekor-tiles/v2/internal/tessera"
+	posixDriver "github.com/sigstore/rekor-tiles/v2/internal/tessera/posix"
+)
+
+var serveCmd = &cobra.Command{
```
cmurphy: It seems like there's some duplication here between this and the other commands. Would it be worthwhile to either 1) extract some of this out into a few shared functions or 2) just make identity mode a flag on the regular posix command instead of a separate command?

internal/server/identity_service.go
loosebazooka: I mean they need to *implement* them to satisfy the grpc def. Unless you mean remove them from the rekor proto def?

cmd/rekor-server/identity-posix/app/serve.go
Hayden-IO: Thanks for the suggestion, having identity server initialization just be behind a flag was much cleaner. Originally I was thinking there was going to be more divergence, but given all of the core logic is housed under `NewServer`, we can just reuse the same command.

Hayden-IO:
Splitting this into two PRs, https://github.com/sigstore/rekor-tiles/pull/871/ with the API and empty server stub first.

internal/server/identity_service.go
Hayden-IO: Removed in https://github.com/sigstore/rekor-tiles/pull/871/

internal/server/identity_service.go
Hayden-IO: No longer an issue! A map of context strings is unneeded, just one.

internal/server/grpc.go
Hayden-IO: I added redundant checks here and in the http server initialization

api/proto/rekor/v2/identity.proto
```diff
@@ -32,6 +32,9 @@ message PublicKeyCredential {
     // The signature computed over the specification identifier, checksum, and context strings,
     // as defined by the c2sp.org/identity-transparency specification.
     bytes signature = 2 [(google.api.field_behavior) = REQUIRED];
+
+    // Optional context value. Must be a SHA-256 digest.
+    bytes context = 3;
```
Hayden-IO, line 38: This is only needed when submitting a public key, so moving under that message.

internal/server/identity_service.go
```diff
@@ -42,8 +53,79 @@ func NewIdentityServer(storage tessera.Storage, readOnly bool, algorithmRegistry
...
+	}
+
+	var extraData []byte
+	var contextBytes []byte
+	if cred, ok := req.Credential.(*pb.IdentityRequestV001_PublicKey); ok {
+		contextBytes = cred.PublicKey.GetContext()
+	}
+
+	if len(contextBytes) > 0 {
+		extraData = fmt.Appendf(nil, "%s:%s", identity.ContextKeyName, hex.EncodeToString(contextBytes))
```
Hayden-IO, line 111: Note that this might change depending on spec discussions.

internal/server/identity_service.go
```diff
@@ -42,8 +53,79 @@ func NewIdentityServer(storage tessera.Storage, readOnly bool, algorithmRegistry
 	}
 }
 
-func (s *IdentityServer) CreateEntry(_ context.Context, _ *pb.IdentityRequestV001) (*httpbody.HttpBody, error) {
-	return nil, status.Errorf(codes.Unimplemented, "method CreateEntry not implemented")
+func (s *IdentityServer) CreateEntry(ctx context.Context, req *pb.IdentityRequestV001) (*httpbody.HttpBody, error) {
+	if s.readOnly {
+		slog.WarnContext(ctx, "rekor is in read-only mode, cannot create new entry")
```
loosebazooka, line 59: Maybe we don't need this anymore, now that we don't have the read APIs in the service definition?

pkg/types/identity/identity.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	switch cred := req.Credential.(type) {
+	case *pb.IdentityRequestV001_PublicKey:
+		if len(cred.PublicKey.GetPublicKey()) == 0 {
+			return errors.New("public key is empty")
+		}
+		if len(cred.PublicKey.GetSignature()) != ed25519.SignatureSize {
+			return errors.New("invalid signature length, must be 64 bytes for Ed25519")
+		}
```
loosebazooka, line 46: Could this be parameterized around an allowed algorithm from the algorithm registry? I don't know exactly what our plan would be for PQ, but I imagine it's a few variations of ml-dsa?

loosebazooka (commented):
looks fine, I was confused because "identity_server" makes it seem like we are focused on the identity case and not the identity+ case. I Don't know if there is a better name though.

pkg/types/identity/identity.go
Hayden-IO: Currently, the algorithm registry doesn't support ML-DSA, because we can't create public APIs using filippo.io/mldsa without introducing a breaking change when we switch to crypto/mldsa. For now, I was planning to just not use the registry and have the server check a) on startup, that the signing key is either ed25519 or ml-dsa-44, and b) per request, the same when a public key is provided as a credential.

We could centralize validation logic under sigstore/sigstore though, either in a utilities library or the registry.

internal/server/identity_service.go
Hayden-IO: I don't recall our exact sharding logic, but keeping this allows us to stop accepting traffic before fully tearing down the log. Even if we don't have read APIs defined for the service, this is still useful.

pkg/types/identity/identity.go
```diff
@@ -0,0 +1,167 @@
...
+	return payload
+}
+
+func computeLeafHash(req *pb.IdentityRequestV001, rootPubKeyHash []byte) []byte {
+	leaf := []byte{LeafVersionV1}
+
+	msgHash := sha256.Sum256(req.GetMessage())
+	leaf = append(leaf, msgHash[:]...)
+
+	leaf = append(leaf, rootPubKeyHash...)
```
cmurphy: Checking this against the example in the spec in https://github.com/C2SP/C2SP/pull/244/changes#diff-4abf7f8874bb57bb178c53a4ad1155e8a4ea89afa6959ed93eedfeee7d7bccacR43-R45 it looks a little backwards, is it 0x01 || root of trust || message or 0x01 || message || root of trust? The example has one order and then the text below it describes something else, probably want to be sure there's agreement on the order before we commit this.

pkg/types/identity/identity.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if len(contextBytes) > 0 {
+		k1 := sha256.Sum256([]byte(ContextKeyName))
+		kDoubleHash := sha256.Sum256(k1[:])
+		vDoubleHash := sha256.Sum256(contextBytes)
+		leaf = append(leaf, kDoubleHash[:]...)
+		leaf = append(leaf, vDoubleHash[:]...)
+	}
+
+	sigHash := sha256.Sum256(sig)
```
cmurphy, line 110: Does the raw sig need to be stored somewhere externally per https://github.com/C2SP/C2SP/pull/244/changes#diff-4abf7f8874bb57bb178c53a4ad1155e8a4ea89afa6959ed93eedfeee7d7bccacR76-R79 ? Or is that up to the client?

api/proto/rekor/v2/identity.proto
cmurphy, line 38: Can you say more about what this means and why only public key needs it? Will this be expanded to include more key-value pairs as described in the spec?

internal/server/grpc.go
```diff
@@ -88,6 +88,13 @@ func newGRPCServer(config *GRPCConfig, server any) *grpcServer {
 
 	s := grpc.NewServer(opts...)
 
+	_, isRekor := server.(pb.RekorServer)
+	_, isIdentity := server.(pb.IdentityRekorServer)
```
cmurphy: The type casting and checking interfaces is a little clunky for Go. Gemini suggests maybe bringing these under a common registration interface like so:

```
    // ServerRegistrar allows different log implementations to register themselves.
    type ServerRegistrar interface {
        RegisterGRPC(s *grpc.Server)
        RegisterHTTP(ctx context.Context, mux *runtime.ServeMux, endpoint string, opts []grpc.DialOption) error
        grpc_health_v1.HealthServer // Ensure they support health check
    }
  
    // internal/server/service.go (Regular Rekor Server):
    func (s *Server) RegisterGRPC(gs *grpc.Server) {
        pb.RegisterRekorServer(gs, s)
    }
    func (s *Server) RegisterHTTP(ctx context.Context, mux *runtime.ServeMux, endpoint string, opts []grpc.DialOption) error {
        return pb.RegisterRekorHandlerFromEndpoint(ctx, mux, endpoint, opts)
    }
  
   // internal/server/identity_service.go (Identity Server):
    func (s *IdentityServer) RegisterGRPC(gs *grpc.Server) {
        pb.RegisterIdentityRekorServer(gs, s)
    }
    func (s *IdentityServer) RegisterHTTP(ctx context.Context, mux *runtime.ServeMux, endpoint string, opts []grpc.DialOption) error {
        return pb.RegisterIdentityRekorHandlerFromEndpoint(ctx, mux, endpoint, opts)
    }
```

then change Serve to have prototype

```
func Serve(ctx context.Context, hc *HTTPConfig, gc *GRPCConfig, timeout time.Duration, reg ServerRegistrar, shutdownFn func(context.Context) error) {
```

What do you think?

identity-posix-compose.yml
```diff
@@ -0,0 +1,95 @@
...
+  rekor:
+    build:
+      context: .
+      target: deploy
+      args:
+        STORAGE_BACKEND: posix
+    command:
+    - "rekor-server"
+    - "serve"
+    - "--identity-mode=true"
```
cmurphy: With the two services being implemented as different types and with two different APIs, it seems a little confusing for it all to be buried in the same command in the same binary under one flag. I think I recall that at one point maybe this had been implemented as a separate command and there was a lot of duplication, and maybe that's why you choose to put it all in the same command under one flag? But I think logically they are still basically different services (with separate docker compose files, even) and maybe belong as separate commands `rekor-server serve` and `rekor-server serve-idenity`. We could extract the common setup code into helper functions to reduce duplication.

tests/identity_e2e_test.go
```diff
@@ -0,0 +1,170 @@
...
+)
+
+var identityPosixConfig = backendConfig{
+	ServerURL:   "http://localhost:3006",
+	StorageURL:  "http://localhost:8001",
+	ComposePath: "identity-posix-compose.yml",
+}
+
+func TestIdentityPOSIX(t *testing.T) {
+	t.Run("ReadWrite", func(t *testing.T) {
```
cmurphy, line 56: Should read-only be tested too?

Hayden-IO: There is no read path, I removed the read APIs since they'd only be implemented for a MySQL-based backend, so I don't think there's anything to test here.

For other tests, deduplication is indirectly tested through the existing e2e tests.

pkg/types/identity/identity.go
Hayden-IO: Good catch. The original version when I wrote the proposal was `0x01 || message || root of trust`, then we flipped it during spec writing - no particular reason why. I wrote this based on an early version of the proposal.

The C2SP PR is the source of truth, it should be flipped here. Will do that now.

pkg/types/identity/identity.go
Hayden-IO: Yes, the log should store this signature as a receipt. I have that as the last TODO on https://github.com/sigstore/rekor-tiles/issues/827. Filippo mentioned an interest in creating a spec for out of tree storage, so I was hoping we could use that. Otherwise, probably will just put the signatures in a bucket, each named by index value, with a TTL of a ~year (TBD, we'll have to set a policy for PGI depending on infra costs and how long we want to be able to prove we received a valid credential).

pkg/types/identity/identity.go
cmurphy: The C2SP PR contradicts itself though (the example versus the text describing it). So it's unclear which version this PR should be aligned with.

api/proto/rekor/v2/identity.proto
Hayden-IO: I'll update the comment on the field to expand on this. This is for uniformity with Sigsum's API, they're making a change to sign over an optional context string. Even if we don't yet have a spec on an upload API, I'd like to try to support the same set of features.

What's a little weird in the spec is that each credential type has a different way of providing context strings. Uploading a signature means providing the context string as well, while an identity token contains the context strings.

I actually originally had this being a map so that it looked more uniform between the two. Arguably, the client could parse the identity token and send the decomposed token as a map of context values along with the identity token signature, though that additional work for clients doesn't seem worth it.

Do you think it'd be useful to have this as a map rather than a single string?

One thing to note is the same of the context key name is TBD, I just called it "context" for now. I thought about not including this at all until the leaf spec is further along, but I thought it might be useful for demos to the transparency community.

pkg/types/identity/identity.go
Hayden-IO: Oh whoops, I'll correct that now. It should be `0x01 || root of trust || message`.

api/proto/rekor/v2/identity.proto
cmurphy: > Do you think it'd be useful to have this as a map rather than a single string?

A map would have been more intuitive to me based on the spec but I'm not fully grasping how this is envisioned to be used so I don't really have an opinion.

Hayden-IO: I think for the sake of having a demo, what we have here is the closest thing to what I believe could be a standardized API. For Sigstore's primary use case, it'll be the log's responsibility to gather the context strings from the identity token, so we might end up just wanting to remove this down the line (or at least flag guard it).

internal/server/grpc.go
Hayden-IO: This looks a lot cleaner. One small thing I did is split the gRPC and HTTP methods into two interfaces, grouping the gRPC server registration and health server registration methods together, with one interface that implements both.

identity-posix-compose.yml
Hayden-IO: How's https://github.com/sigstore/rekor-tiles/pull/828/commits/2c2f110dfd6bcab533bdb474b04ecd7304de2d70 look? This replaces the --identity-mode flag with a separate `serve-identity` command.

api/proto/rekor/v2/identity.proto
```diff
@@ -32,6 +32,10 @@ message PublicKeyCredential {
     // The signature computed over the specification identifier, checksum, and context strings,
     // as defined by the c2sp.org/identity-transparency specification.
     bytes signature = 2 [(google.api.field_behavior) = REQUIRED];
+
+    // Optional context value to record alongside the message. Must be a SHA-256 digest.
+    // Must be signed as defined by the c2sp.org/identity-transparency specification.
+    bytes context = 3;
```
jku, line 38: I don't understand how this matches the C2SP spec proposal, whose job is it to handle the "ecosystem specific" values?

jku: oh, I see, #830 still modifies the proto

loosebazooka:
just gonna rewrite this into 3prs.
- [x] refactor server registration (#974)
- [ ] adjust cmd options
- [ ] core logic of id-based logging

PR #829, opened 2026-06-11
Resolve go fix cleanups, fix e2e tests

Hayden-IO:
Ran go fix ./... on the codebase to modernize the code. This has no functional changes.

This also fixes the end-to-end tests, where a recent change in the witness means we need to add a rate limiter flag.

### DIFF
--- a/internal/server/service_mock.go
+++ b/internal/server/service_mock.go
@@ -196,7 +196,7 @@ func (s *mockRekorServer) GetTile(_ context.Context, in *pb.TileRequest) (*httpb
 	}
 	return &httpbody.HttpBody{
 		ContentType: "application/octet-stream",
-		Data:        []byte(fmt.Sprintf("test-tile:%d,%d,%d", l, i, w)),
+		Data:        fmt.Appendf(nil, "test-tile:%d,%d,%d", l, i, w),
 		Extensions:  nil,
 	}, nil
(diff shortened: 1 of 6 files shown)


internal/server/service_mock.go
```diff
@@ -196,7 +196,7 @@ func (s *mockRekorServer) GetTile(_ context.Context, in *pb.TileRequest) (*httpb
 	}
 	return &httpbody.HttpBody{
 		ContentType: "application/octet-stream",
-		Data:        []byte(fmt.Sprintf("test-tile:%d,%d,%d", l, i, w)),
+		Data:        fmt.Appendf(nil, "test-tile:%d,%d,%d", l, i, w),
```
loosebazooka, line 199: this somehow looks way weirder than the old way?

Hayden-IO: Yea I agree. Unfortunately `go fix` has no way to omit specific lines, you can omit formatters but that would mean all instances of Sprintf->Appendf would be skipped. I'd rather just agree with it.

Hayden-IO: For an AGENTS.md file, I'm planning to specify `go fix` is a requirement for submission, in addition to linters and whatnot

jku (approved):

PR #837, opened 2026-06-18
Add ML-DSA-44 support for public key credential

Hayden-IO:
This introduces support for ML-DSA-44 PQC public keys as per
the identity-based transparency log specification.

The implementation includes an internal package for parsing and
generating ML-DSA public keys per RFC9881. This will be removed once Go
1.27 lands.

Signed-off-by: Hayden <8418760+Hayden-IO@users.noreply.github.com>


internal/crypto/mldsa/mldsa.go
Hayden-IO, line 1: This can now be replaced with filippo.io/mldsa/x509

jku:
Update to golang 1.27 is pretty trivial: https://github.com/jku/rekor-tiles/tree/mldsa-key has a commit

api/proto/rekor/v2/identity.proto
```diff
@@ -26,23 +27,33 @@ option ruby_package = "Sigstore::Rekor::V2";
...
     // The signature computed over the specification identifier, checksum, and context strings,
     // as defined by the c2sp.org/identity-transparency specification.
     bytes signature = 2 [(google.api.field_behavior) = REQUIRED];
+
+    // Optional context value to record alongside the message. Must be a SHA-256 digest.
+    // Must be signed as defined by the c2sp.org/identity-transparency specification.
+    bytes context = 3;
+
+    // The signature algorithm used. Only PKIX_ED25519 and ML_DSA_44 are supported.
+    dev.sigstore.common.v1.PublicKeyDetails algorithm = 4 [(google.api.field_behavior) = REQUIRED];
```
Hayden-IO: Note - get rid of this, SPKI encoding is sufficient

PR #871, opened 2026-07-14
Add Identity-based Transparency Logging API and server stub

Hayden-IO:
This change introduces a new log entry format and dedicated server backend to support Identity-based Transparency Logging as specified in [c2sp.org/identity-transparency](https://github.com/C2SP/C2SP/pull/244).


api/proto/rekor/v2/rekor_identity_service.proto
```diff
@@ -0,0 +1,65 @@
...
+
+// A service for Sigstore clients to log identity usage when signing artifacts
+service IdentityRekor {
+    // Create an identity entry in the log
+    rpc CreateEntry (IdentityRequestV001) returns (google.api.HttpBody) {
+        option (google.api.http) = {
+            post: "/api/v2/log/entries"
+            body: "*"
+        };
+    }
```
Hayden-IO, line 64: Note the lack of read RPCs. Those were originally added I think because we had considered support for a MySQL-only backend, but there's no plans to add that now.

loosebazooka (approved):

PR #872, opened 2026-07-16
Reject pure Ed25519 in hashedrekord instead of panicking

arpitjain099:
### Summary

A hashedrekord request signed with a pure Ed25519 key (`PKIX_ED25519`) crashes `ToLogEntry`. Pure Ed25519 has no prehash, so `algDetails.GetHashType()` returns `crypto.Hash(0)`, and the digest-size check then calls `crypto.Hash(0).Size()`, which panics with "crypto: Size of unknown hash function".

`PKIX_ED25519` is in the default `AllowedClientSigningAlgorithms` set, and the algorithm registry permits an Ed25519 key paired with `Hash(0)`, so a client can trigger this over `CreateEntry`. The gRPC recovery interceptor catches it and returns an internal error per request, but every such request logs a full stack trace and increments the panic counter, so it is a client-reachable per-request crash rather than a purely internal edge case.

### Change

Guard the zero hash before calling `Size()` and return a clear error for a signing algorithm that has no associated digest. Behavior for every other algorithm is unchanged.

### Testing

Added `TestToLogEntryEd25519NoPrehash`, which submits a `PKIX_ED25519` hashedrekord and asserts a clean error. It panics before this change and passes after. `go test ./pkg/types/hashedrekord/ ./internal/server/` is green.

### DIFF
--- a/pkg/types/hashedrekord/hashedrekord.go
+++ b/pkg/types/hashedrekord/hashedrekord.go
@@ -50,6 +50,13 @@ func ToLogEntry(hr *pb.HashedRekordRequestV002, algorithmRegistry *signature.Alg
 	}
 
 	hashAlg := algDetails.GetHashType()
+	// Some signing algorithms (for example pure Ed25519, PKIX_ED25519) have no
+	// associated prehash, so GetHashType returns crypto.Hash(0). A hashedrekord
+	// always carries a message digest, so reject these here instead of calling
+	// Size() on an unset hash, which panics.
+	if hashAlg == crypto.Hash(0) {
+		return nil, fmt.Errorf("unsupported signing algorithm %s for hashedrekord: no associated digest algorithm", hr.Signature.Verifier.KeyDetails)
+	}
 	expectedSize := hashAlg.Size()
 	if len(hr.Digest) != expectedSize {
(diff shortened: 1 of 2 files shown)


pkg/types/hashedrekord/hashedrekord.go
```diff
@@ -50,6 +50,13 @@ func ToLogEntry(hr *pb.HashedRekordRequestV002, algorithmRegistry *signature.Alg
 	}
 
 	hashAlg := algDetails.GetHashType()
+	// Some signing algorithms (for example pure Ed25519, PKIX_ED25519) have no
+	// associated prehash, so GetHashType returns crypto.Hash(0). A hashedrekord
+	// always carries a message digest, so reject these here instead of calling
+	// Size() on an unset hash, which panics.
```
Hayden-IO: Can this comment simply state that only signing algorithms that prehash data are permitted?

arpitjain099: Done, reworded it to lead with the rule: hashedrekord only permits algorithms that prehash the data, with pure Ed25519 as the example that returns crypto.Hash(0).

Hayden-IO (approved):
Thanks!

PR #873, opened 2026-07-16
Remove unused import for deprecated DSSE type

Hayden-IO:
make protos no longer throws a warning


loosebazooka (approved):

loosebazooka:
I think this might be a manual sync back into protobuf specs

PR #902, opened 2026-08-03
Add dispatch tests for the AWS and GCP signer-verifiers

nellshamrell:
## Summary

`internal/tessera/aws/signerverifier` and `internal/tessera/gcp/signerverifier` currently have no test coverage, even though `New()` selects the checkpoint signing key mechanism for every deployment. This adds a test file to each package covering that dispatch. Test-only change.

## Coverage

For each backend:

* **file branch** — returns a usable signer-verifier (sign/verify round trip) and propagates errors for a missing key file
* **KMS branch** — resolves through the provider registry, exercised with the in-memory `fakekms://` provider so no cloud credentials or network access are required (`internal/signerverifier/tink_test.go` already uses this provider)
* **provider registration** — asserts the backend's own scheme (`gcpkms://` / `awskms://`) is registered. This guards the blank import of the production KMS provider: without it, the KMS test above would still pass if that import were dropped. The scheme is written as a literal rather than imported so the assertion depends only on the package's own import. I verified the assertion fails if the blank import is removed.
* **Tink branch** — an unsupported KEK scheme returns `unsupported KMS key type`, and a KEK URI without a keyset path returns `key encryption key URI or keyset path unset`. Both return before any KMS client is constructed, so they stay offline. A successful Tink case is not covered, since it needs a real `gcp-kms://` / `aws-kms://` KEK.

Everything runs offline.

## Notes

* The two packages are near-identical but not copy-pasteable: GCP's `WithKMS` takes `(kms, hash, rpcOpts)` while AWS's takes `(kms, hash)`. If there is interest in deduplicating these backends later, these tests give that refactor a safety net.
* Cases mixing multiple options (e.g. KMS plus file) are intentionally not tested — every caller passes exactly one option via a `switch`, so precedence is an implementation detail rather than intended behavior.
* Verified locally: `go test ./internal/... ./pkg/...`, `go vet`, `golangci-lint run` (v2.12.2, as pinned in `Dockerfile.golangci-lint`), and `addlicense -check`.

### DIFF
--- a/internal/tessera/aws/signerverifier/signerverifier_test.go
+++ b/internal/tessera/aws/signerverifier/signerverifier_test.go
@@ -1,0 +1,141 @@
+/*
+Copyright 2025 The Sigstore Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package signerverifier
+
+import (
+	"bytes"
+	"context"
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"encoding/pem"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/sigstore/sigstore/pkg/signature/kms"
+
+	// fakekms provides an in-memory KMS provider so that the KMS branch of New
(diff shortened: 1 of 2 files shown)


nellshamrell:
This was pushed before it was ready, converting to a draft for a little more review, then will switch back.

nellshamrell:
Ready for review!

internal/tessera/aws/signerverifier/signerverifier_test.go
```diff
@@ -0,0 +1,141 @@
+/*
+Copyright 2025 The Sigstore Authors
```
Hayden-IO: 2025 -> 2026

internal/tessera/gcp/signerverifier/signerverifier_test.go
```diff
@@ -0,0 +1,141 @@
+/*
+Copyright 2025 The Sigstore Authors
```
Hayden-IO: 2025 -> 2026

Hayden-IO (commented):
Thanks, looks great! Just need to update copyright year

nellshamrell:
@Hayden-IO done!

Hayden-IO (approved):

PR #913, opened 2026-08-11
Build with latest version of Go

Hayden-IO:
The version directive in the go.mod file specifies the lowest minor version of Go needed to build the project. Because of this, when we reference the go mod version in the GHA workflows, it will use an outdated version of Go.

This dynamically fetches the Go version from the Dockerfile, which should always use the latest Go version as it's updated by Dependabot. This is the same setup for all the other Sigstore Go projects.

### DIFF
--- a/.github/workflows/license_check.yml
+++ b/.github/workflows/license_check.yml
@@ -27,9 +27,11 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
+      - name: Extract version of Go to use
+        run: echo "GOVERSION=$(awk -F'[:@]' '/FROM golang/{print $2; exit}' Dockerfile)" >> $GITHUB_ENV
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version-file: './go.mod'
+          go-version: ${{ env.GOVERSION }}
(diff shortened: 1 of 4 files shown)


.github/workflows/license_check.yml
```diff
@@ -27,9 +27,11 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
+      - name: Extract version of Go to use
+        run: echo "GOVERSION=$(awk -F'[:@]' '/FROM golang/{print $2; exit}' Dockerfile)" >> $GITHUB_ENV
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version-file: './go.mod'
+          go-version: ${{ env.GOVERSION }}
```
cmurphy, line 34: Would `stable` work here instead? https://github.com/actions/setup-go/blob/main/docs/advanced-usage.md#:~:text=If%20stable%20is%20provided%2C%20the%20action%20will%20get%20the%20latest%20stable%20version%20from%20the%20go%2Dversions%20repository%20manifest.

Hayden-IO: It would, though the downside would be that what we build with and what we test with could diverge, if `stable` is updated before we bump the version in the Dockerfile. This approach gives the same result with every test run.

cmurphy (approved):

PR #925, opened 2026-08-18
Deduplicate the AWS and GCP signer-verifiers

nellshamrell:
## Summary

The per-cloud `signerverifier` packages were near-verbatim copies of each other,
differing only in the KMS import, the KEK URI prefix, the KMS client constructor,
and GCP carrying KMS RPC options. The commands repeated the same bootstrap four
times: `hashAlgMap` was declared verbatim in four files, the file/KMS/Tink switch
in four, and the GCP retry options block in three.

This moves the shared logic into `internal/signerverifier` and command bootstrapping
into `internal/signerverifier/config`.

## Approach

- `internal/signerverifier` holds the signer-verifier construction and the Tink
  keyset handling. It imports no cloud SDKs, so each binary links only the provider
  it actually uses.
- Cloud specifics are inverted into a key encryption key provider that the per-cloud
  package injects through `WithKEKProvider`. Those packages shrink to a KMS import,
  a KEK constructor, and a `New` that supplies it.
- The provider is injected together with the URI prefix it handles, so KEK
  resolution stays **per-cloud** rather than becoming process-global: an AWS binary
  still refuses a `gcp-kms://` key and vice versa, with the same
  `unsupported KMS key type` error as before. The injected option is applied after
  the caller's, so it cannot be overridden.
- `WithKMS` becomes variadic in its RPC options, which lets both clouds share one
  function.
- `internal/signerverifier/config` builds options from the `signer-*` viper keys.
  Its typed errors (`InvalidHashError`, `ErrNoSigner`) let each command keep its
  existing messages: the servers still log the algorithm as a structured attribute
  for an invalid `--signer-kmshash`, and `freeze-checkpoint` still wraps only the
  signer construction failure.

## Behavior

Unchanged. Flags, defaults, error strings, and branch precedence (file → KMS → Tink)
are all preserved, including the ordering that reports an unset keyset path before an
unsupported key type.

## Verification

- `go build ./...` and `go vet ./...` are clean; `go test -race -count=1 ./...` passes.
- Dependency isolation confirmed with `go list -deps` across all five binaries
  (`cmd/rekor-server/{aws,gcp,gcpcloudsql,posix}`, `cmd/freeze-checkpoint/gcp`): the
  AWS binary links neither `tink-go-gcpkms` nor the GCP KMS provider, the GCP
  binaries link neither AWS equivalent, and posix links none of them.
- Per-cloud tests removed from the old packages (`TestNewWithFile`, `TestNewWithKMS`,
  `TestNewErrors`) are relocated to the shared package with equivalent assertions.
  Each cloud package gains tests that its provider is wired, that it rejects the
  other cloud's KEK URI, and that a caller cannot override it.
- The provider-wiring tests pin credential discovery to a missing file, so they
  neither reach the network nor depend on ambient cloud credentials. They were also
  mutation-checked: injecting the wrong prefix makes them fail as intended.


nellshamrell:
Some notes on the design decisions here, to save reviewers time. I'm happy to change any of these — none of them are load-bearing to the goal, and I'd rather match the conventions you prefer.

### Why the per-cloud packages still exist

`internal/tessera/{aws,gcp}/signerverifier` are thin now (~50 lines), so it's fair to ask why they remain rather than folding the wiring into `cmd/`. Two reasons:

- They own the side-effectful blank import (`_ ".../signature/kms/aws"`) that registers the KMS provider. Keeping it in one place per cloud means a single spot to guard with a test.
- Four commands construct signers (three servers plus `freeze-checkpoint`). Inlining the wiring would repeat it in each, which is the duplication this PR is trying to remove.

Worth noting this extends a split that already existed: `internal/signerverifier` was already home to `NewFileSignerVerifier` and `NewTinkSignerVerifierWithHandle`, with the per-cloud packages supplying only the KEK half. This PR moves the remaining cloud-agnostic logic in beside them.

If you'd rather see these collapsed into `cmd/`, I'm glad to do that instead.

### Why the KEK provider is bound to a URI prefix

`WithKEKProvider` takes a prefix alongside the provider, and a Tink KEK URI outside that prefix is rejected before the provider runs. This preserves the pre-refactor property that an AWS-configured binary refuses a `gcp-kms://` key and vice versa, with the same `unsupported KMS key type` error.

The alternative — hand every URI to the provider and let it fail — would mean the URI reaches the cloud SDK before being rejected, which felt like a loosening of the current boundary for no benefit.

### One deliberate but silent behavior

Each wrapper's `New` appends its `WithKEKProvider` option *after* the caller's, so the wrapper's provider wins. That's intentional (a caller shouldn't be able to swap in another cloud's KEK resolver) and there's a test pinning it, but it is silent: a caller-supplied `WithKEKProvider` is ignored rather than rejected.

I went with silent-override to keep `Option` as a plain `func(*config)`, but returning an error there is very reasonable if you'd prefer it to be loud. Happy to change.

### Testing notes

- The provider-wiring tests set `GOOGLE_APPLICATION_CREDENTIALS` to a nonexistent path and disable IMDS for AWS, so credential discovery fails immediately. Without this the GCP test can fall through to a metadata-server probe, which is slow and environment-dependent in CI. The tradeoff is that those tests can't use `t.Parallel`.
- Those tests assert the error is *not* `unsupported KMS key type`, which is a weaker assertion than I'd like. I mutation-checked them: injecting a wrong prefix makes both fail, and they pass again on restore. That isn't visible in the diff, so flagging it here.
- Dependency isolation is verified with `go list -deps` across all five binaries — the AWS binary links neither `tink-go-gcpkms` nor the GCP KMS provider, the GCP binaries link neither AWS equivalent, and posix links none of them. Since the wiring moved from `init()` to explicit injection, this seemed worth re-proving rather than assuming.
- Coverage of the touched packages: `internal/signerverifier` 92.2%, `internal/signerverifier/config` 100%, AWS wrapper 83.3%, GCP wrapper 75.0%.

### Commit structure

This is one squashed commit. If it's easier to review split up, a natural sequence would be: introduce the shared KEK plumbing, migrate AWS, migrate GCP, then extract command bootstrapping into `signerverifier/config`. Just say the word.

### Behavior

No operator-visible change intended: flags, defaults, signer precedence (file → KMS → Tink), and error strings are all preserved, including the ordering that reports an unset keyset path before an unsupported key type. If you spot a case where that isn't true, that's a bug in this PR and I'd like to know.

bobcallaway:
nit: can you please update the copyright year in the new files to be 2026? thx

nellshamrell:
Thanks @bobcallaway! Done — copyright headers in the new files (`internal/signerverifier/{signerverifier.go,kek_test.go}` and `internal/signerverifier/config/{config.go,config_test.go}`) now say 2026. Pushed in a98511b.

PR #926, opened 2026-08-18
Use only one Dockerfile for builds, update Go version

Hayden-IO:
#### Summary

We previously had two separate dockerfiles, one meant for building containers for production and one for per-commit or local builds. These dockerfiles could get out of sync, and had already given the production one was missing a necessary sleep binary for a graceful shutdown.

This PR creates a multi-stage build in a single dockerfile that works for both use cases. Production releases will now target the 'deploy' stage, local builds using docker compose will target the 'local-deploy' stage which is built on a container with shell access, and optionally developers can target the 'debug' stage to get a container with delve. Per-commit builds will target 'local-deploy'.


.github/workflows/build_container.yml
```diff
@@ -92,6 +92,7 @@ jobs:
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
         with:
           context: .
+          target: local-deploy
```
Hayden-IO, line 95: While this could target the debug stage, I thought it wasn't necessary to build delve for the released per-commit containers, so it'll be a slight speedup and reduction in container size.

cmurphy:
I thought we tried having multiple stages in one dockerfile and it didn't work https://github.com/sigstore/rekor-tiles/pull/200#discussion_r2038287896

Hayden-IO:
> I thought we tried having multiple stages in one dockerfile and it didn't work [#200 (comment)](https://github.com/sigstore/rekor-tiles/pull/200#discussion_r2038287896)

Trying to remember exactly what I was thinking when I implemented the early revision of #200, but it looks like I tried to use a conditional to prevent certain steps, and during the later `debug` stage, which would unconditionally run without a target stage, it would fail because the debug binary was missing.

This PR doesn't have that problem because I'm no longer leveraging conditionals to gate steps. Now, building the debug binary has been split into its own stage.

* If you target the `deploy` stage, `docker build --build-arg STORAGE_BACKEND=posix --target deploy . `, it will run up to line 44 building the `builder` and `deploy` stage.
* Targeting `local-deploy`, `docker build --build-arg STORAGE_BACKEND=posix --target local-deploy . `, it will run the `builder` and `local-deploy` stage. It's apparently smart enough to not run the `deploy` stage, because `local-deploy` doesn't reference it.
* Targeting `debug`, `docker build --build-arg STORAGE_BACKEND=posix --target debug . `, it will run the `builder`, `dlvbuilder`, `debug-builder` (using the image from the `builder` stage so it doesn't need to fetch go deps again), and `debug` stages.
* If you run `docker build` with no target stage, it will do the same as above, up to the `debug stage`.

I _think_ my goal was to generate a `deploy` image by default before? IDK, I don't know why I didn't just do this, without any conditionals, as it seems like this works fine.

Hayden-IO:
Also, I might not have realized that you could specify a target build stage in the docker build GHA, and instead tried to control the execution flow with build args.

cmurphy (approved):

PR #927, opened 2026-08-19
Remove pure Ed25519 and restrict allowed signature algorithms

Hayden-IO:
* Removes pure Ed25519 from the default set of allowed client signing algorithms. This was allowed previously because of DSSE but now with only hashedrekord, the pure variant won't work since we don't have the preimage to verify.
* Adds validation in AlgorithmRegistry to prevent the server from specifying algorithms outside the allowed default set, only allowing subsets of the default algorithms. This prevents RSA-PSS and pure Ed25519 from being specified.

### DIFF
--- a/internal/algorithmregistry/algorithmregistry.go
+++ b/internal/algorithmregistry/algorithmregistry.go
@@ -21,6 +21,7 @@ import (
 	"crypto/rsa"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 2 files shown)


Hayden-IO:
Thanks @bobcallaway for finding this!

jku (approved):
lgtm

PR #951, opened 2026-09-04
client: add WithTransport option for a custom base RoundTripper

codysoyland:
The read and write clients build their transport internally and only expose `WithTLSConfig`, so a caller cannot add retries, tracing, or a test double without replacing the whole client. sigstore-go wants to send Rekor v2 writes through sigstore/sigstore's shared `httpretry` transport (sigstore/sigstore-go#676 does this for Fulcio and the TSA; Rekor v2 is waiting on this option), the same way go-tuf's fetcher exposes `SetHTTPClient`. This PR adds `client.WithTransport(http.RoundTripper)` for both `NewReader` and `NewWriter`.

Key changes:
- **`WithTransport` option**: supplies the base `http.RoundTripper`; `CreateRoundTripper` still wraps it to set the User-Agent, so behaviour is unchanged when the option is not used.
- **`Config.BaseTransport()`**: replaces the duplicated transport selection in `NewReader` and `NewWriter`; returns an error when both `WithTransport` and `WithTLSConfig` are set, since a TLS config cannot be applied to an arbitrary RoundTripper.
- **Tests**: option plumbing for reader and writer, the mutual-exclusion error, and an end-to-end `Add` through a counting transport that also checks the User-Agent is applied.

Assisted by Claude Fable 5.1

### DIFF
--- a/pkg/client/read/read.go
+++ b/pkg/client/read/read.go
@@ -62,11 +62,9 @@ func NewReader(readURL, origin string, verifier signature.Verifier, opts ...clie
 	if err != nil {
 		return nil, fmt.Errorf("creating note verifier: %w", err)
 	}
-	transport := http.DefaultTransport
-	if cfg.TLSConfig != nil {
-		transport = &http.Transport{
-			TLSClientConfig: cfg.TLSConfig,
-		}
+	transport, err := cfg.BaseTransport()
+	if err != nil {
+		return nil, err
 	}
 	timeout := cfg.Timeout
 	if timeout == 0 {
--- a/pkg/client/write/write.go
+++ b/pkg/client/write/write.go
@@ -57,11 +57,9 @@ func NewWriter(writeURL string, opts ...client.Option) (Client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("parsing url %s: %w", writeURL, err)
 	}
-	transport := http.DefaultTransport
-	if cfg.TLSConfig != nil {
-		transport = &http.Transport{
-			TLSClientConfig: cfg.TLSConfig,
-		}
+	transport, err := cfg.BaseTransport()
+	if err != nil {
+		return nil, err
 	}
 	timeout := cfg.Timeout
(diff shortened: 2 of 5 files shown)


pkg/client/write/write.go
```diff
@@ -57,11 +57,9 @@ func NewWriter(writeURL string, opts ...client.Option) (Client, error) {
...
 		return nil, fmt.Errorf("parsing url %s: %w", writeURL, err)
 	}
-	transport := http.DefaultTransport
-	if cfg.TLSConfig != nil {
-		transport = &http.Transport{
-			TLSClientConfig: cfg.TLSConfig,
-		}
+	transport, err := cfg.BaseTransport()
+	if err != nil {
+		return nil, err
```
loosebazooka: wrap error?

pkg/client/read/read.go
```diff
@@ -62,11 +62,9 @@ func NewReader(readURL, origin string, verifier signature.Verifier, opts ...clie
...
 		return nil, fmt.Errorf("creating note verifier: %w", err)
 	}
-	transport := http.DefaultTransport
-	if cfg.TLSConfig != nil {
-		transport = &http.Transport{
-			TLSClientConfig: cfg.TLSConfig,
-		}
+	transport, err := cfg.BaseTransport()
+	if err != nil {
+		return nil, err
```
loosebazooka: wrap error?

loosebazooka:
~~I feel like I might be okay with a breaking change here for the rekor-tiles client rather than these runtime checks?~~

This might be a poorer suggestion than I realized if we're trying to be strict about API version

loosebazooka:
hayden is out for a bit, perhaps another go reviewer can take a look at this? @kommendorkapten @cmurphy ?

kommendorkapten (approved):
Looks good!

kommendorkapten:
@loosebazooka I don't have any permissions on this repo but he code looks solid.

loosebazooka (approved):

PR #960, opened 2026-09-07
make idle timeout for http keepalive configurable

bobcallaway:
https://docs.cloud.google.com/load-balancing/docs/https/request-distribution#timeout-keepalive-backends requires idle timeouts to be set > load-balancer value (600s, not configurable) so that the LB controls things. previously this was set to be the same as the global timeout value which was too small.


internal/cli/flags.go
```diff
@@ -38,6 +38,7 @@ func Initialize(serveCmd *cobra.Command) error {
 	serveCmd.Flags().Int("grpc-port", 3001, "GRPC port to bind to")
 	serveCmd.Flags().String("grpc-address", "127.0.0.1", "GRPC address to bind to")
 	serveCmd.Flags().Duration("server-timeout", 20*time.Second, "timeout settings for gRPC and HTTP connections")
+	serveCmd.Flags().Duration("server-idle-timeout", 60*time.Second, "idle timeout settings for gRPC and HTTP connections")
```
loosebazooka, line 41: should this mention that it should be `> load balancer timeout`?

loosebazooka (commented):
I think it mostly seems fine? I assume we set idletimeout to 601 somewhere in our public good config?

gemini seemed to think that setting the idle timeout on the metrics server to match the LB was a little weird... but then you'd have run and plumb a whole new var into this.

jku (approved):

internal/cli/flags.go
bobcallaway: w/ posix there might not be a load balancer

PR #973, opened 2026-09-15
tests: Switch from minio to s3proxy

### LINKED ISSUE #971: Remove test dependency on minio/minio & minio/mc
minio has been adandonware for maybe a year, the repositories were archived a few months ago.

Last week the images disappeared from docker. They're still available on quay.io and can be built... but let's try to find a replacement.
* there is a fork https://github.com/pgsty/silo
* but [s3proxy](https://github.com/gaul/s3proxy) might be an even better fit for our use case

jku:
minio, which we use for local s3 for testing, has been abandoned: #971. Switching is not hugely important but does seem reasonable: use s3proxy instead, it seems a little simpler for our use case.

This is 90% AI generated


aws-compose.yml
```diff
@@ -12,44 +12,32 @@
...
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
-      interval: 5s
-      timeout: 5s
-      retries: 5
+      test:
+        - CMD
+        - bash
+        - -c
+        - exec 3<>"/dev/tcp/127.0.0.1/80" && printf "GET /healthz HTTP/1.0\r\n\r\n" >&3 && head -n 1 <&3 | grep -q " 200 "
```
jku, line 32: the image is lean and does not have curl: use the /dev/tcp bashism as shown in upstream dockerfile

loosebazooka: this seems like a comment we should throw in the yml file?

jku: sure

loosebazooka (dismissed):

loosebazooka (approved):

